PR #12, opened 2023-04-17
[AGENT-4334] Extended CreateProjectOperator to support creating project from dataset_id

oleksandr-saienko:

## Summary
Extended CreateProjectOperator to support creating projects from dataset_id

## Rationale
In most of the cases customers uploading their datasets to AICatalog to reuse it many times.
So its must-have ability to create DR projects from datasets previously uploaded to AICatalog using dataset_id


elatt (approved):

andrius-senulis (changes requested):

andrius-senulis (approved):
looks good now

PR #15, opened 2023-04-18
Fix standalone job def

elatt:

## Summary
- Update standalone job definition
- Add notes about more Jenkins updates

## Rationale
The syntax was incorrect so update job def


dulcardo (approved):

alexromanyuk (approved):
:green_apple:

PR #18, opened 2023-04-24
[AGENT-4349] Implement-UpdateDatasetFromFileOperator

oleksandr-saienko:

## Summary
Added Airflow operator "UpdateDatasetFromFileOperator" that allows to create a new version of existing dataset by uploading a local file.
Also CreateProjectOperator extended to support specific dataset_version_id

## Rationale
We need to allow users to update their existing datasets in AI Catalog with new versions of data

### DIFF
--- a/datarobot_provider/operators/ai_catalog.py
+++ b/datarobot_provider/operators/ai_catalog.py
@@ -59,3 +59,61 @@ class UploadDatasetOperator(BaseOperator):
         ai_catalog_dataset = dr.Dataset.create_from_file(context["params"]["dataset_file_path"])
         self.log.info(f"Dataset created: dataset_id={ai_catalog_dataset.id}")
         return ai_catalog_dataset.id
+
+
+class UpdateDatasetFromFileOperator(BaseOperator):
+    """
+    Operator that creates a new Dataset version from a file.
+    Returns when the new dataset version has been successfully uploaded.
+
+    :param dataset_id: DataRobot AI Catalog dataset ID
+    :type dataset_id: str, optional
+    :param datarobot_conn_id: Connection ID, defaults to `datarobot_default`
+    :type datarobot_conn_id: str, optional
+    :return: DataRobot AI Catalog dataset version ID
+    :rtype: str
+    """
+
+    # Specify the arguments that are allowed to parse with jinja templating
+    template_fields: Iterable[str] = ["dataset_id"]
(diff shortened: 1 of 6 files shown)


oleksandr-saienko:
Jenkins run

datarobot_provider/operators/ai_catalog.py
```diff
@@ -59,3 +59,61 @@ def execute(self, context: Dict[str, Any]) -> str:
...
+    def execute(self, context: Dict[str, Any]) -> str:
+        # Initialize DataRobot client
+        DataRobotHook(datarobot_conn_id=self.datarobot_conn_id).run()
+
+        # If dataset_id not provided in constructor, then using training_dataset_id from json config
+        dataset_id = (
+            self.dataset_id
+            if self.dataset_id is not None
+            else context['params']['training_dataset_id']
+        )
```
elatt, line 107: What is the usecase for these different code paths, e.g. setting the ID from the constructor vs inferring it from the context -- because the file_path is only set from the context so why only the dataset_id can be set in two different ways?

tests/unit/dags/test_create_project_from_dataset_version_dag.py
```diff
@@ -0,0 +1,41 @@
...
+
+from datarobot_provider.example_dags.datarobot_create_project_from_dataset_version_dag import (
+    create_project_from_dataset_version,
+)
+
+
+@pytest.fixture()
+def dagbag(provider_dir):
+    return DagBag(dag_folder=f"{str(provider_dir)}/example_dags", include_examples=False)
+
```
elatt, line 19: This is common in all the unittests right? You can create a `conftest.py` in this directory and move this code here and then the fixture will be reusable across all other files in that directory.

elatt (approved):

dulcardo (approved):

tests/unit/operators/test_ai_catalog.py
```diff
@@ -29,3 +30,27 @@ def test_operator_upload_dataset(mocker):
 
     assert dataset_id == "dataset-id"
     upload_dataset_mock.assert_called_with("/path/to/local/file")
+
+
+def test_operator_update_dataset_from_file(mocker):
```
andrius-senulis, line 35: I'd suggest to parametrize this test to also test the case where operator is initiliazed with  `UpdateDatasetFromFileOperator(dataset_id=<dataset_id>)`.

PR #19, opened 2023-04-28
oleksandr-saienko/AGENT-4243-Implement-JDBC-DataConnectionHook

oleksandr-saienko:

## Summary
Added new connection "DataRobot JDBC DataSource" type that works as a proxy for DataRobot JDBC Connection.
Also Added CreateDatasetFromJDBCOperator to ingest data from JDBC Connection to AI Catalog, and example DAG
## Rationale
Users should be able to configure JDBC Connection using Airflow and use it as a data source with AI Catalog

### DIFF
--- a/datarobot_provider/hooks/connections.py
+++ b/datarobot_provider/hooks/connections.py
@@ -1,0 +1,172 @@
+# Copyright 2022 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from typing import Any
+from typing import Dict
+
+import datarobot as dr
+from airflow import AirflowException
+from airflow.hooks.base import BaseHook
+
+from datarobot_provider.hooks.datarobot import DataRobotHook
+
+
+class JDBCDataSourceHook(BaseHook):
+    """
+    A hook that interacts with DataRobot via its public Python API library.
+
+    :param datarobot_jdbc_conn_id: Connection ID, defaults to `datarobot_jdbc_default`
+    :type datarobot_jdbc_conn_id: str, optional
+    """
+
+    conn_name_attr = 'datarobot_jdbc_conn_id'
+    default_datarobot_jdbc_conn_name = 'datarobot_jdbc_default'
+    conn_type = 'datarobot_jdbc_datasource'
+    hook_name = 'DataRobot JDBC DataSource'
+
+    @staticmethod
+    def get_connection_form_widgets() -> Dict[str, Any]:
+        """Returns connection widgets to add to connection form."""
+        from flask_appbuilder.fieldwidgets import BS3TextFieldWidget
+        from flask_babel import lazy_gettext
+        from wtforms import StringField
+
+        return {
+            "datarobot_connection": StringField(
+                lazy_gettext('DataRobot Connection'),
+                widget=BS3TextFieldWidget(),
+                default='datarobot_default',
+            ),
+            "jdbc_driver": StringField(
+                lazy_gettext('JDBC Driver'),
+                widget=BS3TextFieldWidget(),
+                default='',
+            ),
+            "jdbc_url": StringField(
+                lazy_gettext('JDBC URL'),
+                widget=BS3TextFieldWidget(),
+                default='jdbc:',
+            ),
+        }
+
+    @staticmethod
+    def get_ui_field_behaviour() -> Dict:
+        """Returns custom field behaviour."""
+        return {
+            "hidden_fields": ['host', 'schema', 'port', 'extra'],
+            "relabeling": {},
+            "placeholders": {
+                'datarobot_connection': 'datarobot_default',
+                'jdbc_driver': '',
+                'jdbc_url': 'jdbc:',
+                'login': '',
+                'password': '',
+            },
+        }
+
+    def __init__(
+        self,
+        datarobot_jdbc_conn_id: str = default_datarobot_jdbc_conn_name,
+    ) -> None:
+        super().__init__()
+        self.datarobot_jdbc_conn_id = datarobot_jdbc_conn_id
+
+    def get_conn(self) -> (dict, dr.DataStore):
+        """Initializes a DataRobot DataStore instance with associated credentials."""
+
+        conn = self.get_connection(self.datarobot_jdbc_conn_id)
+
+        datarobot_connection_id = conn.extra_dejson.get('datarobot_connection', '')
+
+        if not datarobot_connection_id:
+            raise AirflowException("datarobot_connection is not defined")
+
+        # Initialize DataRobot client by DataRobotHook
+        DataRobotHook(datarobot_conn_id=datarobot_connection_id).run()
+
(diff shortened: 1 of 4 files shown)


oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,182 @@
...
+        # For methods that accept credential data instead of credential ID
+        credential_data = {
+            "credentialType": "basic",
+            "user": conn.login,
+            "password": conn.password,
+        }
+
+        # Find the JDBC driver ID from name:
+        listm = DataDriver.list()
+        print(listm)
```
andrius-senulis, line 118: this

tests/unit/conftest.py
```diff
@@ -36,3 +36,62 @@ def mock_airflow_connection(mocker, dr_conn_details):
...
+    return {
+        "login": "test_login",
+        "password": "test_password",
+        "datarobot_connection": "datarobot_default",
+        "jdbc_driver": "test JDBC Driver",
+        "jdbc_url": "jdbc:test_jdbc_connection_string",
+    }
+
+
+@pytest.fixture(scope="function", autouse=True)
```
andrius-senulis, line 53: nit: `scope="function"` is the default, so no need to mention it explicitely

andrius-senulis (approved):
LGTM

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,182 @@
...
+    default_datarobot_jdbc_conn_name = 'datarobot_jdbc_default'
+    conn_type = 'datarobot_jdbc_datasource'
+    hook_name = 'DataRobot JDBC DataSource'
+
+    @staticmethod
+    def get_connection_form_widgets() -> Dict[str, Any]:
+        """Returns connection widgets to add to connection form."""
+        from flask_appbuilder.fieldwidgets import BS3TextFieldWidget
+        from flask_babel import lazy_gettext
+        from wtforms import StringField
```
elatt, line 38: Why are you importing these here and not at the top of the module?

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,182 @@
...
+            ),
+            "jdbc_driver": StringField(
+                lazy_gettext('JDBC Driver'),
+                widget=BS3TextFieldWidget(),
+                default='',
+            ),
+            "jdbc_url": StringField(
+                lazy_gettext('JDBC URL'),
+                widget=BS3TextFieldWidget(),
+                default='jdbc:',
```
elatt, line 54: Is it worth having defaults that don't work or can we have None, as default? Does this display in the UI when the user first opens it?

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,182 @@
...
+        """Returns custom field behaviour."""
+        return {
+            "hidden_fields": ['host', 'schema', 'port', 'extra'],
+            "relabeling": {},
+            "placeholders": {
+                'datarobot_connection': 'datarobot_default',
+                'jdbc_driver': '',
+                'jdbc_url': 'jdbc:',
+                'login': '',
+                'password': '',
```
elatt, line 69: Do all jdbc connections have a username and password? I thought you can encode username and password values directly in the connection URL?

datarobot_provider/hooks/connections.py
elatt, line 118:
```suggestion
```

datarobot_provider/operators/ai_catalog.py
```diff
@@ -117,3 +118,89 @@ def execute(self, context: Dict[str, Any]) -> str:
...
+            params.data_store_id = data_store.id
+            # Creating DataSource using params with data_store_id
+            self.log.info(f"Creating DataSource: {dataset_name}")
+            data_source = dr.DataSource.create(
+                data_source_type='jdbc', canonical_name=dataset_name, params=params
+            )
+            self.log.info(f"DataSource:{dataset_name} successfully created, id={data_source.id}")
+
+        # Checking if there are any changes in params:
+        elif not (params == data_source.params):
```
elatt, line 192:
```suggestion
        elif params != data_source.params:
```

datarobot_provider/hooks/connections.py
oleksandr-saienko: thats because of DataRobot API, its expecting username and password as a separate credentials_data object. Also its not secure to keep password as a part of connection string and sometimes it might cause problems with special characters inside a password. So I would suggest to force users to use login&password as a separate fields

oleksandr-saienko:
Jenkins run

tests/unit/conftest.py
elatt:
```suggestion
@pytest.fixture(autouse=True)
```

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,180 @@
...
+        jdbc_url = conn.extra_dejson.get('jdbc_url', '')
+
+        if not jdbc_url:
+            raise AirflowException("jdbc_url is not defined")
+
+        if not conn.login:
+            raise AirflowException("login is not defined")
+
+        if not conn.password:
+            raise AirflowException("password is not defined")
```
alexromanyuk, line 107: Can we have a configuration as dictionary with Python's standard data types? Then some [JSON Schema validator](https://pypi.org/project/jsonschema/) can be used to specify required fields and any additional rules.

alexromanyuk (approved):

datarobot_provider/hooks/connections.py
```diff
@@ -0,0 +1,180 @@
...
+        self,
+        datarobot_jdbc_conn_id: str = default_datarobot_jdbc_conn_name,
+    ) -> None:
+        super().__init__()
+        self.datarobot_jdbc_conn_id = datarobot_jdbc_conn_id
+
+    def get_conn(self) -> (dict, DataStore):
+        """Initializes a DataRobot DataStore instance with associated credentials."""
+
+        conn = self.get_connection(self.datarobot_jdbc_conn_id)
```
elatt, line 83: What's the difference between `get_conn` and `get_connection`? Should `get_conn` simply override `get_connection` and call its super class rather than have a unique method? It also looks like it is doing more than just getting a connection (i.e. `DataStore.create`) so maybe it should be named `find_or_create_datastore` ?

datarobot_provider/operators/ai_catalog.py
```diff
@@ -117,3 +118,89 @@ def execute(self, context: Dict[str, Any]) -> str:
...
+
+        dataset_name = context["params"]["dataset_name"]
+
+        data_source = None
+
+        for dr_source_item in dr.DataSource.list():
+            if dr_source_item.canonical_name == dataset_name:
+                data_source = dr_source_item
+                self.log.info(f"Found existing DataSource:{dataset_name}, id={data_source.id}")
+                break
```
elatt, line 169: Minor pythonic nit, but this is the perfect case for Python's `for-else` support:
```python
for dr_source_item in dr.DataSource.list():
    if dr_source_item.canonical_name == dataset_name:
        data_source = dr_source_item
        break
else:
    # No existing datasource was found so create one
    data_source = dr.DataSource.create(...)
```

PR #20, opened 2023-05-04
[AGENT-4434] Fix for error handling for case of using wrong JDBC Driver name

oleksandr-saienko:

## Summary
Fixed error handling for case where user added wrong JDBC Driver name

## Rationale
User should get proper error message in case of wrong JDBC Driver name


oleksandr-saienko:
Jenkins run

elatt (approved):

PR #22, opened 2023-05-10
[AGENT-4242] add gcp credentials management from airflow

oleksandr-saienko:

## Summary
Added support for GCP Credentials, extended GetCredentialIdOperator to support fetching existing Credentials or replicating Credentials based on stored Airflow connection 
added unit-tests

## Rationale
Users should be able to create GCP credentials form Airflow, and use it for scoring and project creation

### DIFF
--- a/datarobot_provider/hooks/credentials.py
+++ b/datarobot_provider/hooks/credentials.py
@@ -6,9 +6,10 @@
 #
 # Released under the terms of DataRobot Tool and Utility Agreement.
 
+import json
 from typing import Any
 from typing import Dict
-import json
+
 from airflow import AirflowException
 from airflow.hooks.base import BaseHook
 from datarobot.models.credential import Credential
@@ -216,8 +217,8 @@ class GoogleCloudCredentialsHook(CredentialsBaseHook):
     @staticmethod
     def get_connection_form_widgets() -> Dict[str, Any]:
         """Returns connection widgets to add to connection form."""
-        from flask_appbuilder.fieldwidgets import BS3TextFieldWidget
         from flask_appbuilder.fieldwidgets import BS3TextAreaFieldWidget
+        from flask_appbuilder.fieldwidgets import BS3TextFieldWidget
         from flask_babel import lazy_gettext
         from wtforms import StringField
 
--- a/datarobot_provider/hooks/credentials.py
+++ b/datarobot_provider/hooks/credentials.py
@@ -109,7 +109,7 @@ class CredentialsBaseHook(BaseHook):
 
 class BasicCredentialsHook(CredentialsBaseHook):
     hook_name = 'DataRobot Basic Credentials'
-    conn_type = 'datarobot-basic-credentials'
+    conn_type = 'datarobotcredentialsbasic'
 
     def __init__(
         self,
@@ -176,7 +176,7 @@ class BasicCredentialsHook(CredentialsBaseHook):
 
 class GoogleCloudCredentialsHook(CredentialsBaseHook):
     hook_name = 'DataRobot GCP Credentials'
-    conn_type = 'datarobot_gcp_credentials'
+    conn_type = 'datarobotcredentialsgcp'
 
     def create_credentials(self, conn) -> Credential:
         """Returns Google Cloud credentials for params in connection object"""
--- a/datarobot_provider/hooks/credentials.py
+++ b/datarobot_provider/hooks/credentials.py
@@ -8,7 +8,7 @@
 
 from typing import Any
 from typing import Dict
-
+import json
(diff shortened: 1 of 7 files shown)


datarobot_provider/hooks/credentials.py
```diff
@@ -61,7 +62,7 @@ def get_conn(self) -> Credential:
         # Trying to find existing DataRobot Credentials managed by Airflow provider:
```
elatt, line 62: While you are in this method, can you rename it please `get_conn` is ambiguous and confusing with how it differs from `self.get_connection`. Maybe it should be called `get_or_create_credential`?

tests/unit/conftest.py
```diff
@@ -138,3 +147,34 @@ def mock_airflow_connection_datarobot_basic_credentials(mocker, dr_basic_credent
         ),
     )
     mocker.patch.dict("os.environ", AIRFLOW_CONN_DATAROBOT_BASIC_CREDENTIALS_DEFAULT=conn.get_uri())
+
+
+@pytest.fixture(scope="function", autouse=True)
```
elatt, line 152:
```suggestion
@pytest.fixture(autouse=True)
```

`scope="function"` is the default so you don't need to specify that.

tests/unit/conftest.py
```diff
@@ -138,3 +147,34 @@ def mock_airflow_connection_datarobot_basic_credentials(mocker, dr_basic_credent
...
+    )
+
+    mocker.patch("datarobot_provider.hooks.credentials.Credential.list", return_value=[])
+    mocker.patch(
+        "datarobot_provider.hooks.credentials.Credential.create_gcp",
+        return_value=gcp_credentials_create_mock,
+    )
+
+
+@pytest.fixture(scope="function", autouse=True)
```
elatt, line 169:
```suggestion
@pytest.fixture(autouse=True)
```

tests/unit/conftest.py
elatt: You also have a lot of fixtures that have `autouse=True`. That's typically a code smell and having them here in such a high level `conftest.py` can end up being very surprising to test writers in the future (e.g. methods are patched that they didn't expect to be w/o having read through the entire contents of this file to understand all the places that are mocked out).

Are these fixtures really so essential or can they only be active in specific tests?

datarobot_provider/hooks/credentials.py
oleksandr-saienko: get_conn is implementation of get_conn abstract method from BaseHook:     
```
def get_conn(self) -> Any:
        """Returns connection for the hook."""
        raise NotImplementedError()
```

oleksandr-saienko: Propose for  get_connection is different is just fetching stored values from secret storage
    def get_connection(cls, conn_id: str) -> Connection:
        conn = Connection.get_connection_from_secrets(conn_id)

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

datarobot_provider/hooks/credentials.py
elatt: Ok, then if we are stuck with `get_conn`, I feel we should rename `get_connection`

oleksandr-saienko: get_connection is also BaseHook method, so we cant rename it, its core class of Airflow

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

datarobot_provider/hooks/credentials.py
elatt: Ah ok, that's unfortunate then

oleksandr-saienko:
Jenkins run

dulcardo (approved):
LGTM

elatt (approved):
Giving a 👍 on this PR under the presumption we will cleanup the `autouse` stuff in a follow up. I'm happy to chat if you want to 🧠 ⛈️ .   My thoughts are we either just declare required fixtures on the tests themselves that require them (explicit approach). Or, if we think that will be too onerous because we expect the GCP credential mock to be used on several tests that have to do with GCP, maybe we can reorg the files a little to be more credential type focused so we can use autouse in a more targeted manner.

PR #24, opened 2023-05-15
[AGENT-4240] Add AWS credentials support

oleksandr-saienko:

## Summary
Added support for storing and using AWS credentials on Airflow side

## Rationale
Users should be able to create DataRobot credentials for AWS using  Airflow UI

### DIFF
--- a/datarobot_provider/hooks/credentials.py
+++ b/datarobot_provider/hooks/credentials.py
@@ -245,3 +245,95 @@ class GoogleCloudCredentialsHook(CredentialsBaseHook):
                 'gcp_key': 'Enter a valid JSON string',
             },
         }
+
+
+class AwsCredentialsHook(CredentialsBaseHook):
+    hook_name = 'DataRobot AWS Credentials'
+    conn_type = 'datarobot.credentials.aws'
+
+    def create_credentials(self, conn) -> Credential:
+        """Returns AWS credentials for params in connection object"""
+
+        aws_access_key_id = conn.extra_dejson.get('aws_access_key_id', '')
+
+        if not aws_access_key_id:
+            raise AirflowException("aws_access_key_id is not defined")
+
+        aws_secret_access_key = conn.extra_dejson.get('aws_secret_access_key', '')
+
+        if not aws_secret_access_key:
+            raise AirflowException("aws_secret_access_key is not defined")
+
+        # aws_session_token is optional:
+        aws_session_token = conn.extra_dejson.get('aws_session_token', None)
+
+        try:
+            self.log.info(f"Creating AWS Credentials:{self.datarobot_credentials_conn_id}")
+            credential = Credential.create_s3(
+                name=self.datarobot_credentials_conn_id,
+                aws_access_key_id=aws_access_key_id,
+                aws_secret_access_key=aws_secret_access_key,
+                aws_session_token=aws_session_token,
+                description=self.default_credential_description,
+            )
+
+            return credential
+
+        except Exception as e:
+            self.log.error(
+                f"Error creating AWS Credentials: {self.datarobot_credentials_conn_id}, message:{str(e)}"
+            )
+            raise AirflowException(
+                f"Error creating AWS Credentials: {self.datarobot_credentials_conn_id}"
+            )
+
+    def get_credential_data(self, conn) -> dict:
+        # For methods that accept credential data instead of credential ID
+        credential_data = {
+            "credentialType": "s3",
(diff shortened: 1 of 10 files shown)


oleksandr-saienko:
Jenkins run

datarobot_provider/hooks/credentials.py
```diff
@@ -245,3 +245,95 @@ def get_ui_field_behaviour() -> Dict:
...
+                widget=BS3TextFieldWidget(),
+                default='datarobot_default',
+            ),
+            'aws_access_key_id': StringField(
+                lazy_gettext("AWS access key ID"),
+                widget=BS3TextFieldWidget(),
+            ),
+            'aws_secret_access_key': StringField(
+                lazy_gettext("AWS secret access key"),
+                widget=BS3TextFieldWidget(),
```
elatt, line 324: Shouldn't this be a password so it is masked?

tests/unit/conftest.py
```diff
@@ -116,14 +116,12 @@ def mock_airflow_connection_datarobot_jdbc(mocker, dr_jdbc_conn_details):
...
 
-@pytest.fixture(autouse=True)
+@pytest.fixture()
 def mock_datarobot_basic_credentials(mocker, dr_basic_credentials_conn_details):
     credentials_create_mock = mocker.Mock(
         credential_id='test-credentials-id',
         name='datarobot_basic_credentials_test',
         credential_type='basic',
-        user=dr_basic_credentials_conn_details["login"],
-        password=dr_basic_credentials_conn_details["password"],
```
elatt, line 126: Why is the username and password being deleted from the mock?

tests/unit/conftest.py
```diff
@@ -149,13 +149,12 @@ def mock_airflow_connection_datarobot_basic_credentials(mocker, dr_basic_credent
...
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture()
 def mock_datarobot_gcp_credentials(mocker, dr_gcp_credentials_conn_details):
     gcp_credentials_create_mock = mocker.Mock(
         credential_id='test-gcp-credentials-id',
         name='datarobot_gcp_credentials_test',
         credential_type='gcp',
-        gcp_key=dr_gcp_credentials_conn_details["gcp_key"],
```
elatt, line 158: ditto, why is the secret being deleted?

tests/unit/conftest.py
```diff
@@ -178,3 +179,41 @@ def mock_airflow_connection_datarobot_gcp_credentials(mocker, dr_gcp_credentials
...
+
+
+@pytest.fixture()
+def mock_datarobot_aws_credentials(mocker, dr_aws_credentials_conn_details):
+    aws_credentials_create_mock = mocker.Mock(
+        credential_id='test-aws-credentials-id',
+        name='datarobot_aws_credentials_test',
+        credential_type='s3',
+        description="Credentials managed by Airflow provider for Datarobot",
+    )
```
elatt, line 202: Why doesn't this contain the actual secret values?

tests/unit/conftest.py
```diff
@@ -116,14 +116,12 @@ def mock_airflow_connection_datarobot_jdbc(mocker, dr_jdbc_conn_details):
     mocker.patch.dict("os.environ", AIRFLOW_CONN_DATAROBOT_JDBC_DEFAULT=conn.get_uri())
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture()
 def mock_datarobot_basic_credentials(mocker, dr_basic_credentials_conn_details):
```
elatt, line 120:
```suggestion
def mock_datarobot_basic_credentials(mocker):
```

It doesn't look like you use this fixture here anymore?

tests/unit/conftest.py
oleksandr-saienko: its because its mock for Credential object which doesn't contain login/password

tests/unit/conftest.py
oleksandr-saienko: it was previously added by mistake, its mock for Credential object which doesn't contain anything except credential_id, name, credential_type and description

tests/unit/conftest.py
oleksandr-saienko: its a mock for the output of Credential.create_s3 method which doesn't contain anything except credential_id, name, credential_type and description

elatt (approved):

tests/unit/conftest.py
elatt: Ah that's right, the secret data isn't actually contained in a DR API `Credential`.

dulcardo (approved):

PR #28, opened 2023-05-19
[AGENT-4245] Implement CreateDataStoreOperator

oleksandr-saienko:

## Summary
Added CreateDataStoreOperator that returned datastore_id by connection name. It creates new datastore using preconfigured jdbc connection in case if connection not exist

## Rationale
users should be able to fetch datastore_id for existing connections because it required for different operators

### DIFF
--- a/datarobot_provider/operators/connections.py
+++ b/datarobot_provider/operators/connections.py
@@ -1,0 +1,77 @@
+# Copyright 2023 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from typing import Any
+from typing import Dict
+from typing import Iterable
+
+from airflow.exceptions import AirflowException
+from airflow.exceptions import AirflowNotFoundException
+from airflow.models import BaseOperator
+
+from datarobot_provider.hooks.connections import JDBCDataSourceHook
+from datarobot_provider.hooks.datarobot import DataRobotHook
+
+DATAROBOT_MAX_WAIT = 3600
+DATAROBOT_AUTOPILOT_TIMEOUT = 86400
+DATETIME_FORMAT = "%Y-%m-%d %H:%M:%s"
+
+
+class GetOrCreateDataStoreOperator(BaseOperator):
+    """
+    Fetching DataStore by connection name or creating if it does not exist
+    and return DataStore ID.
+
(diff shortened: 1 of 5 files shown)


oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

dulcardo (approved):

datarobot_provider/operators/connections.py
```diff
@@ -0,0 +1,77 @@
...
+
+from airflow.exceptions import AirflowException
+from airflow.exceptions import AirflowNotFoundException
+from airflow.models import BaseOperator
+
+from datarobot_provider.hooks.connections import JDBCDataSourceHook
+from datarobot_provider.hooks.datarobot import DataRobotHook
+
+DATAROBOT_MAX_WAIT = 3600
+DATAROBOT_AUTOPILOT_TIMEOUT = 86400
```
elatt, line 20: It's always good to include in the variable name the unit for these types of things, my guess is:
```suggestion
DATAROBOT_MAX_WAIT_SEC = 3600
DATAROBOT_AUTOPILOT_TIMEOUT_SEC = 86400
```

datarobot_provider/operators/connections.py
```diff
@@ -0,0 +1,77 @@
...
+    :type datarobot_conn_id: str, optional
+    :return: DataRobot Credentials ID
+    :rtype: str
+    """
+
+    # Specify the arguments that are allowed to parse with jinja templating
+    template_fields: Iterable[str] = []
+    template_fields_renderers: Dict[str, str] = {}
+    template_ext: Iterable[str] = ()
+    ui_color = '#f4a460'
```
elatt, line 41:
```suggestion
    ui_color = '#f4a460'  # SandyBrown
```

datarobot_provider/operators/connections.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Fetch stored JDBC Connection with credentials
+        credential_data, data_store = JDBCDataSourceHook(
+            datarobot_jdbc_conn_id=connection_name
+        ).run()
+
+        if data_store is not None:
+            self.log.info(f"Found preconfigured jdbc connection: {connection_name}")
+
+        return data_store.id
```
elatt, line 77: This code doesn't look like it handles the case where `data_store` can be `None`. Or the value can never be none which would make the log statement irrelevant.

elatt (approved):
Other than a few small things, looks good.

PR #33, opened 2023-05-30
[AGENT-4488] Implement Airflow Operator for reading Deployment metrics

oleksandr-saienko:

## Summary
Implemented GetServiceStatsOperator to read Deployment metrics and serialize it to pass to the next operator with user logic. Added example DAG with GetServiceStatsOperator and custom Python operator template

## Rationale
Users should be able to read service metrics from Deployment to use it with their custom logic

### DIFF
--- a/datarobot_provider/operators/monitoring.py
+++ b/datarobot_provider/operators/monitoring.py
@@ -1,0 +1,72 @@
+# Copyright 2023 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from typing import Any
+from typing import Dict
+from typing import Iterable
+from typing import List
+
+import datarobot as dr
+from airflow.exceptions import AirflowException
+from airflow.models import BaseOperator
+
+from datarobot_provider.hooks.datarobot import DataRobotHook
+
+DATETIME_FORMAT = "%Y-%m-%d %H:%M:%s"
+
+
+class GetServiceStatsOperator(BaseOperator):
+    """
+    Gets service stats measurements from a deployment.
+
+    :param deployment_id: DataRobot deployment ID
+    :type deployment_id: str
+    :param datarobot_conn_id: Connection ID, defaults to `datarobot_default`
+    :type datarobot_conn_id: str, optional
+    :return: Service stats for a Deployment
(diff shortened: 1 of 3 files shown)


oleksandr-saienko:
Jenkins run

datarobot_provider/operators/monitoring.py
```diff
@@ -0,0 +1,72 @@
...
+class GetServiceStatsOperator(BaseOperator):
+    """
+    Gets service stats measurements from a deployment.
+
+    :param deployment_id: DataRobot deployment ID
+    :type deployment_id: str
+    :param datarobot_conn_id: Connection ID, defaults to `datarobot_default`
+    :type datarobot_conn_id: str, optional
+    :return: Service stats for a Deployment
+    :rtype: List[dict]
```
elatt, line 31: Is it possible to show even more of an example of what kind of key/values are in the dict? Or point them to our public API : https://docs.datarobot.com/en/docs/api/reference/public-api/deployments.html#example-responses_29

datarobot_provider/operators/monitoring.py
```diff
@@ -0,0 +1,72 @@
...
+        service_stats = deployment.get_service_stats(**service_stats_params)
+        return _serialize_service_stats(service_stats)
+
+
+def _serialize_service_stats(service_stats_obj, date_format=DATETIME_FORMAT):
+    service_stats_dict = service_stats_obj.__dict__.copy()
+    service_stats_dict["period"] = {
+        "start": service_stats_obj.period["start"].strftime(date_format),
+        "end": service_stats_obj.period["end"].strftime(date_format),
+    }
```
elatt, line 71: Rather than having our own specific format, I'd personally prefer outputting as ISO format (https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat)

elatt (approved):
LGTM but I personally think it would be better if we stringified the datetime object as ISO format.

datarobot_provider/operators/monitoring.py
oleksandr-saienko: This information will be added in scope of "Update documentation" as the final task of current milestone

datarobot_provider/operators/monitoring.py
oleksandr-saienko: output of this operator can be used as an input for other operator, so we should keep date format same for all operators

alexromanyuk (approved):

PR #36, opened 2023-06-07
[AGENT-4486] Implement Airflow Operators to create new dataset versions from datasource

oleksandr-saienko:

## Summary
Added Airflow Operators: CreateDatasetVersionOperator and CreateDatasetFromDataStoreOperator to support creating new versions of existing datasets
changed "schedule_interval" to "schedule" for all DAGs because "schedule_interval" was deprecated

## Rationale
Users should be able to create new versions of existing datasets in AI Catalog. Also it required for submit actuals DAG because "dynamic" dataset is not supported for submit actuals method

### DIFF
--- a/datarobot_provider/example_dags/datarobot_dataset_new_version_dag.py
+++ b/datarobot_provider/example_dags/datarobot_dataset_new_version_dag.py
@@ -37,7 +37,7 @@ from datarobot_provider.operators.credentials import GetOrCreateCredentialOperat
     },
 )
 def datarobot_dataset_new_version(
-    deployment_id='646fcfe9b01540a797f224b3', dataset_id='647f3d5ec731b1fd66eadb29'
+    deployment_id=None, dataset_id=None
 ):
     if not deployment_id:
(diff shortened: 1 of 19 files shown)


datarobot_provider/example_dags/datarobot_dataset_new_version_dag.py
```diff
@@ -0,0 +1,73 @@
...
+    # Default json config example:
+    params={
+        "datarobot_jdbc_connection": "datarobot_jdbc_demo",
+        "dataset_name": "integration_example_demo",
+        "table_schema": "integration_example_demo",
+        "table_name": "actuals_demo",
+    },
+)
+def datarobot_dataset_new_version(
+    deployment_id='646fcfe9b01540a797f224b3', dataset_id='647f3d5ec731b1fd66eadb29'
```
elatt, line 40:
```suggestion
    deployment_id=None, dataset_id=None
```

elatt (approved):

PR #42, opened 2023-06-27
[AGENT-4613] Add Airflow DAG predictions examples with Scoring Code

oleksandr-saienko:

## Summary
Added Airflow DAG predictions examples with Scoring Code
Added Operators for scoring code downloading from deployment and from model

## Rationale
Users should be able to use Airflow with scoring code to run predictions

### DIFF
--- a/datarobot_provider/example_dags/download_scoring_code_from_deployment_dag.py
+++ b/datarobot_provider/example_dags/download_scoring_code_from_deployment_dag.py
@@ -1,0 +1,74 @@
+# Copyright 2023 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from datetime import datetime
+
+from airflow.decorators import dag
+from airflow.decorators import task
+
+from datarobot_provider.operators.scoring_code import DownloadDeploymentScoringCodeOperator
+
+
+@dag(
+    schedule=None,
+    start_date=datetime(2023, 1, 1),
+    tags=['example', 'scoring_code'],
+    # Default json config example:
+    params={
+        'source_code': False,
+        'include_agent': False,
+        'include_prediction_explanations': False,
+        'include_prediction_intervals': False,
+        # as an option you can pass base_path here:
+        # 'scoring_code_filepath': "/home/airflow/gcs/data/",
+    },
+)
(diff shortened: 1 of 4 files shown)


oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

datarobot_provider/example_dags/download_scoring_code_from_deployment_dag.py
```diff
@@ -0,0 +1,74 @@
...
+        # The next scoring code example is only for demo and testing proposes,
+        # should be used only with a very small datasets (not higher than a few megabytes).
+        # To run this example, DataRobot Prediction Library (https://datarobot.github.io/datarobot-predict)
+        # should be installed first.
+
+        input_file = "/.../input_file.csv"
+        output_file = "/.../output_predictions.csv"
+
+        import pandas as pd
+        from datarobot_predict.scoring_code import ScoringCodeModel
```
andrius-senulis, line 57: why do you need to have imports in the function rather than top of module?

tests/unit/operators/test_scoring_code.py
```diff
@@ -0,0 +1,178 @@
...
+
+    mocker.patch.object(dr.Deployment, "download_scoring_code")
+
+    operator = DownloadDeploymentScoringCodeOperator(
+        task_id="download_scoring_code_from_deployment",
+        deployment_id="deployment-id",
+        base_path=scoring_code_filepath,
+    )
+
+    with pytest.raises(ValueError):
```
andrius-senulis, line 78: you may also check that the right error is raised with:
```
with pytest.raises(ValueError, match="Invalid or missing base path value"):
```

tests/unit/operators/test_scoring_code.py
andrius-senulis, line 1: I would also test that it works as expected when `DownloadDeploymentScoringCodeOperator`  is called with `base_path=None` but `context["params"].get("scoring_code_filepath")` is set. Also for model operator.

andrius-senulis (approved):
A few minor comments.

datarobot_provider/example_dags/download_scoring_code_from_deployment_dag.py
oleksandr-saienko: its just a sample DAG, and datarobot_predict currently is not part of airflow provider so I decided to put all required stuff related to scoring inside Python operator

PR #55, opened 2023-08-04
[AGENT-4682] Implement Airflow Operators for Training a specific blueprint 

oleksandr-saienko:

## Summary
Added TrainModelOperator and ModelTrainingJobSensor for model training using specific blueprint
## Rationale
users should be able to train a model using specific blueprint

### DIFF
--- a/datarobot_provider/sensors/model_insights.py
+++ b/datarobot_provider/sensors/model_insights.py
@@ -9,6 +9,7 @@ from typing import Any
 from typing import Dict
 
 from airflow.sensors.base import BaseSensorOperator
+from airflow.sensors.base import PokeReturnValue
 from datarobot import Job
 from datarobot.errors import AsyncProcessUnsuccessfulError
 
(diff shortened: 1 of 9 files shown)


oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

oleksandr-saienko:
Jenkins run

dulcardo (approved):

datarobot_provider/sensors/model_insights.py
```diff
@@ -47,7 +49,11 @@ def __init__(
 
         self.hook = DataRobotHook(datarobot_conn_id)
 
-    def poke(self, context: Dict[Any, Any]) -> bool:
+    def get_job_result(self, context: Dict[Any, Any]) -> bool:
+        # Default implementation return True if job is completed:
+        return True
```
elatt, line 54: Should this be abstract? I'm not familiar enough with the DR api to know all the nuances but it seems weird to have a default implementation that just returns true?

elatt (approved):

datarobot_provider/sensors/model_insights.py
oleksandr-saienko: All good, some jobs dont need return any object, just job complete status, so we can use DataRobotJobSensor directly (we just need return True when job is complete), but some jobs returns a result object after job is complete in this case we need to use corresponding child class (like ModelTrainingJobSensor this case) and return PokeReturnValue which contains job status and some job result object (in this case PokeReturnValue(True, trained_model.id))

PR #60, opened 2023-08-28
[AGENT-4712] Implement Airflow Operator for Execution Environment Creation

oleksandr-saienko:

## Summary
Implemented CreateExecutionEnvironmentOperator to create Custom Execution Environment

## Rationale
Users should be able to create Custom Execution Environment using Airflow operators

### DIFF
--- a/datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
+++ b/datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
@@ -1,0 +1,38 @@
+# Copyright 2023 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+
+from datetime import datetime
+
+from airflow.decorators import dag
+
+from datarobot_provider.operators.execution_environment import CreateExecutionEnvironmentOperator
+
+
+@dag(
+    schedule=None,
+    start_date=datetime(2022, 1, 1),
+    tags=['example', 'custom model'],
+    params={
+        "execution_environment_name": "Demo Execution Environment",
+        "execution_environment_description": "Demo Execution Environment for Airflow provider",
+        "programming_language": "python",
+        "required_metadata_keys": [{"field_name": "test_key", "display_name": "test_display_name"}],
+    },
+)
+def create_execution_environment():
+    create_execution_environment_op = CreateExecutionEnvironmentOperator(
+        task_id='create_execution_environment',
+    )
+
+    create_execution_environment_op
+
+
+create_execution_environment_dag = create_execution_environment()
+
+if __name__ == "__main__":
+    create_execution_environment_dag.test()
--- a/datarobot_provider/operators/execution_environment.py
+++ b/datarobot_provider/operators/execution_environment.py
@@ -1,0 +1,94 @@
+# Copyright 2023 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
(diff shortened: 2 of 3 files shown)


datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@dag(
+    schedule=None,
+    start_date=datetime(2022, 1, 1),
+    tags=['example', 'custom model'],
+    params={
+        "execution_environment_name": "Demo Execution Environment",
+        "execution_environment_description": "Demo Execution Environment for Airflow provider",
+        "programming_language": "python",
```
elatt, line 23: Should this be an enum from the datarobot client lib?

datarobot_provider/operators/execution_environment.py
```diff
@@ -0,0 +1,94 @@
...
+        if required_metadata_keys:
+            metadata_keys = [
+                RequiredMetadataKey(field_name=key["field_name"], display_name=key["display_name"])
+                for key in required_metadata_keys
+            ]
+
+        execution_environment = dr.ExecutionEnvironment.create(
+            name=self.name
+            if self.name is not None
+            else context["params"].get("execution_environment_name", None),
```
elatt, line 80: I didn't think these expressions could span multiple lines, at least not without using ()'s

I think it would be easier to read if you pulled them out since they are so long.

elatt (approved):
Some minor comments but otherwise looks good

datarobot_provider/operators/execution_environment.py
oleksandr-saienko: its after "black" reformatting, will try to make it more readable

datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
oleksandr-saienko: similar enum exist for custom tasks: CUSTOM_TASK_LANGUAGE = enum(PYTHON="python", R="r", JAVA="java", OTHER="other"), but original "create Execution Environment" method from datarobot client lib didn't refer to this enum, mention just this string can be “python”, “r”, “java” or “other”

alexromanyuk (approved):

datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
elatt: I wouldn't use the custom tasks enum because they may be similar now but aren't really related so no guarantees PEs will match all values in the future.

elatt: Also, fwiw, we accept snowflake and azure now too.

oleksandr-saienko: yes, thats why Im not using enum here, so probably better to keep string value as in original "create Execution Environment" method from datarobot client lib, but datarobot client lib should be refactored first in this case

oleksandr-saienko: I guess for "snowflake and azure" you are referring to External Prediction Environment for external deployments, but not Custom Execution Environment for custom models

oleksandr-saienko: Or is it related to "Proxy models"?

elatt: Ah sorry, I thought this stuff was dealing with Prediction envs not custom models stuff.

PR #62, opened 2023-08-29
[AGENT-4710] Implement Airflow Operator to create an Execution Environment Version

oleksandr-saienko:

## Summary
implemented CreateExecutionEnvironmentVersionOperator to create new version of Execution Environment
added unit-tests
extended sample DAG

## Rationale
Users should be able to create Execution Environment Version using Airflow operators

### DIFF
--- a/datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
+++ b/datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
@@ -10,8 +10,10 @@ from datetime import datetime
 
 from airflow.decorators import dag
 
-from datarobot_provider.operators.execution_environment import CreateExecutionEnvironmentOperator, \
-    CreateExecutionEnvironmentVersionOperator
+from datarobot_provider.operators.execution_environment import CreateExecutionEnvironmentOperator
+from datarobot_provider.operators.execution_environment import (
+    CreateExecutionEnvironmentVersionOperator,
+)
 
 
 @dag(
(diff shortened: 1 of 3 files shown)


oleksandr-saienko:
Jenkins run

datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
```diff
@@ -11,25 +11,35 @@
...
     schedule=None,
-    start_date=datetime(2022, 1, 1),
+    start_date=datetime(2023, 1, 1),
     tags=['example', 'custom model'],
     params={
         "execution_environment_name": "Demo Execution Environment",
         "execution_environment_description": "Demo Execution Environment for Airflow provider",
         "programming_language": "python",
         "required_metadata_keys": [{"field_name": "test_key", "display_name": "test_display_name"}],
+        "docker_context_path": "./datarobot-user-models-master/",
```
elatt, line 28: How does this work in a cloud managed Airflow env? Do those platforms allow you to mount files from some other storage and it can show up as a local path?

elatt (approved):

datarobot_provider/example_dags/datarobot_create_execution_environment_dag.py
oleksandr-saienko: yes, for cloud managed Airflow there is a automatic mapping to cloud storage to use it as a local path

PR #87, opened 2024-03-28
[AGENT-5518] Bump airflow-provider-datarobot version to 0.0.9 to test Harness CICD

oleksandr-saienko:

## Summary
Bump airflow-provider-datarobot version to 0.0.9 to test Harness CICD

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 apache-airflow>=2.3.0
 black>=22.12.0
-datarobot>=3.3.0
+datarobot>=3.3.1
 flake8>=4.0.1
 isort>=5.11.4
 mypy>=0.931
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ with open("README.md", "r") as fh:
 """Perform the package airflow-provider-datarobot setup."""
 setup(
     name='airflow-provider-datarobot',
-    version="0.0.8",
+    version="0.0.9",
     description='DataRobot Airflow provider.',
     long_description=long_description,
     long_description_content_type='text/markdown',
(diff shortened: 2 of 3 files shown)


datarobot-devinfragit6:
pipelineController framework is deprecated, please migrate CICD for this repository to Harness

alexromanyuk (approved):

oleksandr-saienko:
Jenkins run

dulcardo (approved):

setup.py
```diff
@@ -35,7 +35,7 @@
         'datarobot_provider.sensors',
         'datarobot_provider.operators',
     ],
-    install_requires=['apache-airflow>=2.3.0', 'datarobot>=3.3.0'],
+    install_requires=['apache-airflow>=2.3.0', 'datarobot>=3.3.1'],
```
elatt, line 38: Do you really need the new version? There were no code changes so why force users to use newer version of library?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 apache-airflow>=2.3.0
```
elatt, line 1: You should really be pinning exact versions (e.g. ==) in requirements file so we have repeatable tests.

setup.py
oleksandr-saienko: Bumped minimal datarobot version to 3.3.1 because it contains bugfixes

elatt: You should try and keep as low a version as possible, it makes it easier for this package to be installed. Unless if this library requires those bug fixes.

requirements.txt
oleksandr-saienko: in most of the cases user can already have apache-airflow in place so we shouldn't limiting this

elatt: This file should be only used for our internal testing so we should be explicit about what versions we pull down so our tests are repeatible. In `setup.py` (or pyproject.toml) we specify ranges of versions of packages we support, in this case we should try and be as broad as possible and have as open of ranges as possible to not create dependency conflicting scenarios for customers.

elatt: I'm not saying we need to do it now but at some point in the future.

elatt (approved):
Looks ok; a few minor suggestions.

PR #88, opened 2024-04-02
[AGENT-5162] Migrate pipeline controller cicd pipelines for repository airflow provider datarobot to harness (JUST FOR TESTING, DO NOT MERGE IT)

oleksandr-saienko:
THIS IS FOR TESTING ONLY
will be deleted after

### DIFF
--- a/.harness/black-pylint-on-PR.yaml
+++ b/.harness/black-pylint-on-PR.yaml
@@ -1,0 +1,44 @@
+pipeline:
+  name: black-pylint-on-PR
+  identifier: black_check
+  projectIdentifier: dummy
+  orgIdentifier: AGENTS
+  tags: {}
+  properties:
+    ci:
+      codebase:
+        connectorRef: account.svc_harness_git1
+        repoName: airflow-provider-datarobot
+        build: <+input>
+  stages:
+    - parallel:
+        - stage:
+            name: black
+            identifier: black
+            description: ""
+            type: CI
+            spec:
+              cloneCodebase: true
+              execution:
+                steps:
+                  - step:
+                      type: Run
+                      name: black check step
+                      identifier: Run_1
+                      spec:
+                        connectorRef: account.dockerhub_datarobot_read
+                        image: python:3.9
+                        shell: Bash
+                        command: |-
+                          set -exuo pipefail
+                          pip install -U pip
+                          pip install -r requirements.txt
+                          make lint check-licenses
+              infrastructure:
+                type: KubernetesDirect
+                spec:
+                  connectorRef: account.cigeneral
+                  namespace: harness-delegate-ng
+                  automountServiceAccountToken: true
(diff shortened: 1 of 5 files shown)


elatt:
Can you fill out PR description please

.harness/black-pylint-on-PR.yaml
```diff
@@ -0,0 +1,45 @@
+pipeline:
+  name: black-pylint-on-PR
+  identifier: black_check
+  projectIdentifier: dummy
```
elatt, line 4: I think we need a real project now

.harness/buildandpublishpipy.yaml
```diff
@@ -0,0 +1,71 @@
+pipeline:
+  name: build-and-publish-pipy
+  identifier: buildandpublishpipy
+  projectIdentifier: dummy
```
elatt, line 4: Ditto

.harness/buildandpublishpipy.yaml
```diff
@@ -0,0 +1,71 @@
...
+    ci:
+      codebase:
+        connectorRef: account.svc_harness_git1
+        repoName: airflow-provider-datarobot
+        build: <+input>
+  stages:
+    - parallel:
+        - stage:
+            name: black
+            identifier: black
```
elatt, line 17: I think this needs to be renamed to match what it is actually doing.

.harness/dummytestpipeline.yaml
```diff
@@ -0,0 +1,83 @@
+pipeline:
+  name: dummy-test-pipeline
```
elatt, line 2: Is this really a dummy pipeline? It looks useful and different from the simple lint pipeline. Can we just simplify for now and combine the pipeline? I.e. no need to run tests if lint is failing.

.harness/default_input_set.yaml
```diff
@@ -0,0 +1,14 @@
+inputSet:
+  name: default_input_set
```
elatt, line 2: Does this have to be its own file or can it be included in the file it is associated with? Would be easier to manage if it could.

build_publish_early_access.yaml
```diff
@@ -0,0 +1,71 @@
...
+                      spec:
+                        envVariables:
+                          PYPI_TOKEN: <+secrets.getValue("pypi-airflow-provider-datarobot-early-access")>
+                        connectorRef: account.dockerhub_datarobot_read
+                        image: python:3.9
+                        shell: Bash
+                        command: |-
+                          set -exuo pipefail
+                          pip install --upgrade twine
+                          echo "Publishing wheel to pip..."
```
elatt, line 63: It must be subtle because I'm missing how this is different from the other publish?

elatt:
Ah, didn't realize this was for testing because I got a notice that you requested my review

PR #95, opened 2024-04-12
[AGENT-5619] Add Harness pipeline to publish airflow-provider-datarobot-early-access

oleksandr-saienko:

## Summary
Added Harness pipeline to publish airflow-provider-datarobot-early-access to PyPi


elatt (commented):
How does this pipeline work? It looks like the same logic as the official publish -- how does it publish the early access version of the package and not the official one, doesn't it have to change the name if the package and build a new wheel?

alexromanyuk (approved):

mjnitz02:
Closing this as we're working on a new set of pipelines to mirror the pysdk

PR #100, opened 2025-01-14
[MLDEV-1124] Update codeowners to have a more domain-oriented review process

mjnitz02:

## Summary
This is a rough first draft at attempting to allocate the review into the appropriate slack channels so it doesn't always ping the 3 people listed in the codeowners file. It could be good to improve this by adding in some linter automation for the codeowners file (e.g. `code-owners format -i` as supported in the monolith) and some checks related to this. First I just want to at least get it where the reviews are being spread out to the domains/individuals working on things.

I think that `machine-learning-development` is probably an ok owner of the base files, and `continuous-integration` can be tagged on harness related changes. Actual test/dag changes and other areas, we'll need to sort of identify as this project progresses and we can dynamically improve on the codeowners file.

## Rationale
- Make codeowners point at domains

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,14 @@
-*       @elatt @alexromanyuk @andrius-senulis
+/                                                        @datarobot/core-backend
+/.harness                                                @datarobot/continuous-integration
+/datarobot_provider                                      @datarobot/machine-learning-development
+/datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
+/datarobot_provider/operators/connections.py             @datarobot/mldata
+/datarobot_provider/operators/feature_discovery.py       @datarobot/trust-explainable-ai
+/datarobot_provider/operators/mlops.py                   @datarobot/model-management
+/datarobot_provider/operators/model_insights.py          @datarobot/trust-explainable-ai
+/datarobot_provider/operators/model_package.py           @datarobot/model-management
+/datarobot_provider/operators/model_predictions.py       @datarobot/predictions
+/datarobot_provider/operators/monitoring.py              @datarobot/model-management
+/datarobot_provider/operators/monitoring_job.py          @datarobot/model-management
+/datarobot_provider/operators/prediction_explanations.py @datarobot/trust-explainable-ai
+/tests/                                                  @datarobot/machine-learning-development
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
-/                                                        @datarobot/machine-learning-development
+*                                                        @datarobot/machine-learning-development
 /.harness                                                @datarobot/continuous-integration
 /datarobot_provider                                      @datarobot/machine-learning-development
 /datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
-/                                                        @datarobot/core-backend
+/                                                        @datarobot/machine-learning-development
 /.harness                                                @datarobot/continuous-integration
 /datarobot_provider                                      @datarobot/machine-learning-development
 /datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,12 +1,12 @@
 *                                                        @datarobot/machine-learning-development
 /datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
 /datarobot_provider/operators/connections.py             @datarobot/mldata
-/datarobot_provider/operators/feature_discovery.py       @datarobot/trust-explainable-ai
+/datarobot_provider/operators/feature_discovery.py       @datarobot/mldata
 /datarobot_provider/operators/mlops.py                   @datarobot/model-management
 /datarobot_provider/operators/model_insights.py          @datarobot/trust-explainable-ai
 /datarobot_provider/operators/model_package.py           @datarobot/model-management
 /datarobot_provider/operators/model_predictions.py       @datarobot/predictions
 /datarobot_provider/operators/monitoring.py              @datarobot/model-management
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1 +1,14 @@
-*       @elatt @alexromanyuk @andrius-senulis
+*                                                        @datarobot/machine-learning-development
```
mjnitz02, line 1: It shows `CODEOWNERS` errors here, but the bots resolve the tags properly, so I think this shouldn't be an issue.

CODEOWNERS
```diff
@@ -1 +1,14 @@
...
+/.harness                                                @datarobot/continuous-integration
```
elatt, line 2: The CI team is going to own the pipelines in this repo? I thought typically they just own the infra, not individual repo pipelines? cc @datarobot/continuous-integration

mjnitz02: I guess I can move all this under `machine-learning-dev` for now. I saw that they owned a bunch of `.harness` folders in places, and didn't know if they wanted to review this stuff.

CODEOWNERS
elatt: As long as our bots still work fine, I suppose that's ok. I imagine GitHub complains because this repo is public but the groups you are listing are private 🤷‍♂️

mjnitz02: @elatt Yea this is the problem, it says the groups listed don't resolve to publicly known groups

CODEOWNERS
```diff
@@ -1 +1,12 @@
-*       @elatt @alexromanyuk @andrius-senulis
+*                                                        @datarobot/machine-learning-development
+/datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
+/datarobot_provider/operators/connections.py             @datarobot/mldata
+/datarobot_provider/operators/feature_discovery.py       @datarobot/trust-explainable-ai
```
bdrosen96, line 4: Should feature discovery also be mldata?

mbromanowsky (approved):
I agree that Feature Discovery should be MLData, but otherwise LGTM for MLDev and TREX.

In the monorepo, our code owners file is called DRCODEOWNERS -- is that to avoid the issue of non-public team handles? Could we point our bots at a DRCODEOWNERS here too?

In our public-facing packages on PyPI, we have basically a mailing list behind a shell account called `datarobot` for the primary owner. Would that trick help here?

elatt (approved):
I'm ok with these changes but added the other teams for review so they can be aware of this change.

CODEOWNERS
```diff
@@ -1 +1,12 @@
...
+/datarobot_provider/operators/mlops.py                   @datarobot/model-management
+/datarobot_provider/operators/model_insights.py          @datarobot/trust-explainable-ai
+/datarobot_provider/operators/model_package.py           @datarobot/model-management
+/datarobot_provider/operators/model_predictions.py       @datarobot/predictions
+/datarobot_provider/operators/monitoring.py              @datarobot/model-management
+/datarobot_provider/operators/monitoring_job.py          @datarobot/model-management
```
automactic, line 10: `monitoring_job.py` should be owned by tracking agent

CODEOWNERS
```diff
@@ -1 +1,12 @@
-*       @elatt @alexromanyuk @andrius-senulis
+*                                                        @datarobot/machine-learning-development
+/datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
+/datarobot_provider/operators/connections.py             @datarobot/mldata
+/datarobot_provider/operators/feature_discovery.py       @datarobot/trust-explainable-ai
+/datarobot_provider/operators/mlops.py                   @datarobot/model-management
+/datarobot_provider/operators/model_insights.py          @datarobot/trust-explainable-ai
+/datarobot_provider/operators/model_package.py           @datarobot/model-management
+/datarobot_provider/operators/model_predictions.py       @datarobot/predictions
+/datarobot_provider/operators/monitoring.py              @datarobot/model-management
```
automactic, line 9: FYI we will merge `mlops.py` and `monitoring.py` into `deployment.py`, so these two files will be deleted soon

mjnitz02:
> I agree that Feature Discovery should be MLData, but otherwise LGTM for MLDev and TREX.
> 
> In the monorepo, our code owners file is called DRCODEOWNERS -- is that to avoid the issue of non-public team handles? Could we point our bots at a DRCODEOWNERS here too?
> 
> In our public-facing packages on PyPI, we have basically a mailing list behind a shell account called `datarobot` for the primary owner. Would that trick help here?

@mbromanowsky I'm going to try to sync up with DSX to see if I can get some help improving this since its a public facing repo. The bots will use either `CODEOWNERS` or `DRCODEOWNERS` depending on which exists, so they are pretty agnostic to this situation. The rest of it is more resolving what is/isn't available to this repo from a public/private perspective. I'm hoping DSX can maybe help me with a bit of the admin related cleanup of this repo.

CODEOWNERS
```diff
@@ -1 +1,12 @@
-*       @elatt @alexromanyuk @andrius-senulis
+*                                                        @datarobot/machine-learning-development
+/datarobot_provider/operators/ai_catalog.py              @datarobot/mldata
+/datarobot_provider/operators/connections.py             @datarobot/mldata
+/datarobot_provider/operators/feature_discovery.py       @datarobot/mldata
```
mjnitz02, line 4: @mbromanowsky Updated this to mldata

CODEOWNERS
mjnitz02: Updated this so its now fixed.

bdrosen96 (approved):

mjnitz02:
@mbromanowsky Maybe so, I can play around with that idea and see what happens :D

andrius-senulis (approved):

CODEOWNERS
tsdaemon: Should it be `/ @datarobot/machine-learning-development`? I suppose `*` means that mldev will be tagged for ALL changes

PR #101, opened 2025-01-15
[MLDEV-1126] Improve makefile commands for astro dev environments

mjnitz02:

## Summary
This PR reworks some of the `Makefile` to make the development flow much smoother and quicker. We don't include the `airflow_provider_datarobot` as a package directly anymore and just point at a second `requirements_dev.txt` that is used after the initial image is built, but before the airflow environment is started. The `Makefile` will take care of adjusting the `Dockerfile` and `requirements_dev.txt` files during rebuilds.

The reason for using this slightly offset approach is that the `Docker` environment inside astro runs a few different commands, and the files are not present inside the container at the stage where the `requirements.txt` is read and executed. We need to wait until the initial container is constructed and execute a second `pip install` command between the initial build, and the airflow environment actually executing.

The makefile basically shortens the build flow by:
- Stop running airflow environments in astro
- Cleaning out the `dist` folder
- Building the current package to a wheel
- Updating the `requirements_dev.txt` with the wheel location directly (we don't use a `==` here because we just want to directly install whatever is in the wheel. This avoids having to bump/change versions during development)
- Restarting and rebuilding the astro environment

_This process takes ~10-30 seconds from start to finish, so it's pretty fast to keep dev times easy_

I also added helper files with OS detection for:
- Installing astro
- Creating a dev environment
- Starting/stopping astro
- Rebuilding the environment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,18 @@
 SHELL := bash
 
+# Determine the operating system
+# "darwin" or "linux"
+OS_NAME := $(shell uname -s | tr A-Z a-z)
+ifeq ($(OS_NAME), darwin)
+	ASTRO_CMD=brew install astro
+else
+	ASTRO_CMD=curl -sSL install.astronomer.io | sudo bash -s
+endif
+
 .PHONY: black isort lint typecheck check-licenses fix-licenses unit-tests
+
+req:
+	pip install -r requirements.txt
 
 lint:
 	flake8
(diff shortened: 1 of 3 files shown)


mjnitz02:
cc @andrius-senulis  @tsdaemon Here is a PR to help out with the usability. The basic flow for a new dev environment would be:
```
make reset-astro-dev
make build-astro-dev
```

and on changes you want to test you just run
`make build-astro-dev`

I've added a few additional helpers here, but this should keep everything contained in a single `git-ignored` repo and make it easy and fast

Makefile
```diff
@@ -24,3 +36,34 @@ fix-licenses:
...
 check-licenses:
 	docker run  --rm -v $(CURDIR):/github/workspace ghcr.io/apache/skywalking-eyes/license-eye:785bb7f3810572d6912666b4f64bad28e4360799 -v info -c .licenserc.yaml header check
+
+install-astro:
+	$(ASTRO_CMD)
+
+create-astro-dev:
+	mkdir astro-dev
+	cd astro-dev && astro dev init
+	echo "RUN pip install -r \"/usr/local/airflow/requirements_dev.txt\"" >> ./astro-dev/Dockerfile
```
elatt, line 46: Should you make this idempotent? E.g. only edit the file if we don't see this entry already present?

Makefile
```diff
@@ -24,3 +36,34 @@ fix-licenses:
...
+	$(MAKE) create-astro-dev
+
+start-astro-dev:
+	cd astro-dev && astro dev start
+
+stop-astro-dev:
+	cd astro-dev && astro dev stop
+
+build-astro-dev:
+	echo "For testing remember to bump datarobot_provider/__init__.py and setup.py versions"
```
elatt, line 59:
```suggestion
	@echo "For testing remember to bump datarobot_provider/__init__.py and setup.py versions"
```

mjnitz02: @elatt Thanks for pointing this out, let me remove this. I made a fix in the latest commit that gets rid of the need for this. So its no longer necessary to bump any versions for testing purposes.

Makefile
mjnitz02: I can, this command will only work on building a new environment once. The `astro dev init` command would always fail unless its a new astro environment. So this would only be runnable once because of `Line 45`

PR #102, opened 2025-01-16
Do not copy examples by default

tsdaemon:
Two minor changes:
* Add `-` to stop commands so downstream is still executed when stop fails. And stop fails often when its not run.
* Do not copy examples by default: I really feel like its unnecessary side-effect of build


mjnitz02 (approved):
Thanks for these changes. I didn't know about the `-` operator :D

PR #104, opened 2025-01-17
[MLDEV-1132] Use ruff instead of isort/black/flake8/lint

tsdaemon:

## Summary

* Replace a zoo of static code checking tools with a single entrypoint.
* Add a harness step to verify formatting
* Reformat everything: I have selected `quote-style = "double"` as it generated less changes. I saw that originally we used `skip-string-normalization = true`: I don't think there is any convention behind it, IMO we just have a mix of single and double quotes everywhere.

## Rationale

Ruff is better because its:
* all-in-one batteries-included tool which covers flake8, lint, black, isort, and any other static code checking tools (including [Aiflow static code checks](https://docs.astral.sh/ruff/rules/#airflow-air)).
* faster then all of those because its written in rust
* single entry point for configuration

### DIFF
--- a/datarobot_provider/hooks/credentials.py
+++ b/datarobot_provider/hooks/credentials.py
@@ -7,7 +7,8 @@
 # Released under the terms of DataRobot Tool and Utility Agreement.
 
 import json
-from typing import Any, Dict
+from typing import Any
+from typing import Dict
 
 from airflow import AirflowException
 from airflow.hooks.base import BaseHook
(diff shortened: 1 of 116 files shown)


nek-taras (approved):

elatt (approved):

andrius-senulis (approved):
Awesome!

rabih-datarobot (approved):
LG for TREX

datarobot_provider/hooks/credentials.py
```diff
@@ -166,7 +166,7 @@ def update_credentials(self, conn, credential: Credential) -> None:
             )
             raise AirflowException(
                 f"Error updating Basic Credentials: {self.datarobot_credentials_conn_id}"
-            )
+            ) from None
```
jpclemens0, line 169: What does this do?

jpclemens0 (approved):
LGTM for MLDev

datarobot_provider/hooks/credentials.py
tsdaemon: It makes exception self contained, otherwise it will be shown as "During handling this exception another exception occured". Lint recommendation

mjnitz02 (approved):
Saw the JIRA ticket this mornign and spend some time reviewing ruff. I wasn't aware of it, but it looks like a really good all-in-one solution thats faster and more direct

PR #106, opened 2025-01-17
[MLDEV-1136] Get basic sphinx documentation up and running

mjnitz02:

## Summary
This PR largely pulls from the basic build methodology of the `public_python_client` with a bit of slimming down.  Most of this code is a direct transplant of the `sdk_docs` concept and helpers.

It adds support for:
- Homepage
- Installation
- Connections
- a few basic initial classes

We will need to expand on this and make sure all of the docs are working properly, but this at least gets an initial working implementation running.

A harness pipeline also needs to be stood up for testing the documentation, but that will be done in a followup PR.

I've also added a `dev` variant of setup (matches how the python client works.

## Rationale
We need to have some docs for this repo. This will need to be done over several stages, but this at least gets a very basic implementation of some of the APIs and a foundation up and running. The `make testdocs` command completes successfully on this repo.


mjnitz02:
@tsdaemon I removed the `requirements.txt` from this repo and replaced it with a setuptools `dev` install. This makes it function more in line with the other public repos.

mjnitz02:
@tsdaemon Never mind, that will require some pipeline updates, so that can be a follow up

docs/make.bat
```diff
@@ -0,0 +1,190 @@
+@ECHO OFF
```
tsdaemon, line 1: Why do we need a make.bat? I suppose one would need it to build docs on Windows, is this something we need?

mjnitz02: Let me remove it for now, it might be a leftover that the sdk never removed. I don't think we need it and we can always readd it

setup.py
```diff
@@ -35,10 +53,13 @@
         "datarobot_provider.sensors",
         "datarobot_provider.operators",
     ],
-    install_requires=["apache-airflow>=2.3.0", "datarobot>=3.3.1"],
+    install_requires=["apache-airflow>=2.3.0", "datarobot>=3.6.0"],
```
tsdaemon, line 56: I think this deserves a new version

mjnitz02: I can drop this back to 3.3.1 for now. I think we could bump it when we move to `py 3.12`.

tsdaemon: I don't mind this upgrade. Mind that I don't think migration to `py 3.12` means that we will only support `py 3.12` for this library: I believe we should support all alive version of Python. Its just that for development its better to target the latest python.

tsdaemon (approved):
LGTM

ahjota:
> This PR largely pulls from the basic build methodology of the public_python_client with a bit of slimming down. Most of this code is a direct transplant of the sdk_docs concept and helpers.

For what it's worth, CFX is planning on dropping nearly all of this in the next few months. We're moving to MkDocs for better integration with the docs portal.

mjnitz02:
@ahjota Ah, that works, we can probably move this over to `MkDocs` also then. I just wanted to start getting something up so we can also start cleaning up the readme and checking the docstrings are actually legit. I think the docstrings are going to need a lot of tweaks/improvements looking at the current builds

docs/conf.py
```diff
@@ -0,0 +1,408 @@
...
+# Adapted from https://github.com/liran-funaro/sphinx-markdown-builder?tab=readme-ov-file#configurations
+
+# If set to True, then anchors will be added before each section/function/class signature.
+# This allows references to a specific anchor in the document.
+# markdown_anchor_sections: True
+# markdown_anchor_signatures: True
+
+# If set to True, adds metadata to the top of each document containing author, copyright,
+# and version.
+markdown_docinfo: True
```
c-h-russell-walker, line 281: Hey 👋 

So while working on an (unsolicited) PR for mypy updates - [seen here](https://github.com/datarobot/airflow-provider-datarobot/pull/109/) - I now see a mypy error that _could_ signify this should be an assignment like so:
`markdown_docinfo = True`

Since I see the error:
```
docs/conf.py:281:19: error: Invalid type: try using Literal[True] instead?  [valid-type]
    markdown_docinfo: True
```

So mypy wants the typing to be updated which makes me think that maybe the syntax is off here - even though the comments above _do_ suggest using a colon.

Anyway, just wanted to share - _hopefully_ helpful 🤷

PR #109, opened 2025-01-18
[-] Update mypy config and fix some low hanging fruit errors/issues

c-h-russell-walker:
# This repository is public. Do not put here any private DataRobot or customer's data: code, datasets, model artifacts, .etc.

## Summary

Update mypy config to have some important rules.
(I left `warn_return_any` commented out since it increases the number of errors)

Update as many errors/issues as I can while not being a code-owner of affected code. (Soon I will likely be a contributor on behalf of the Notebooks team to create Airflow operators for use with our DR Notebooks)

@mjnitz02 & @andrius-senulis hopefully this unsolicited PR is helpful. I am likely going to contribute soon and wanted to see the state of typing in this repo. and see if I could help out at all.

I pulled in the two commits from this PR https://github.com/datarobot/airflow-provider-datarobot/pull/108 for my usage of the pipe operator to not break the tests (that PR ups the version of python used in pipelines).

## Rationale

As we move forward on adding more to this repository I believe having type checking in place for at the very least all _new_ code is important.

Ideally it would be nice to have mypy run as part of PR checks and any pipelines so that we and external users can confidently rely on type annotations.

## Related

I took the python version of `3.11` (as updated in the mypy config file) from this PR:
https://github.com/datarobot/airflow-provider-datarobot/pull/108

## Screenshots

_Errors on current main branch (before PR) are:_
`Found 285 errors in 31 files (checked 127 source files)`

Screenshot 2025-01-18 at 11 13 54 AM

<hr>

_Errors when run with this PR's code:_
`Found 50 errors in 18 files (checked 126 source files)`

Screenshot 2025-01-19 at 8 49 03 AM

PR #110, opened 2025-01-18
[-] Get mypy errors to zero in order to enforce that all new code must pass mypy linting

c-h-russell-walker:

## Summary

This work is dependent on previous PR here:
https://github.com/datarobot/airflow-provider-datarobot/pull/109

## Rationale

I believe it's important that we can enforce mypy during the lint pipeline.

What this means we've done in this PR is:
1. Consolidate to use pyproject.toml (instead of mypy.ini
2. Simplify make command
3. Update lint pipeline to _also_ run mypy checks
4. And importantly add a `tool.mypy.overrides` in the mypy's config section to ignore existing errors

What is important is we _hopefully_ prioritize actually fixing the errors as seen in those listed modules (maybe per domain potentially) so we can remove those lines from the config and also that will mean moving forward those modules will then (after that fact) have to pass mypy linting.

<hr>

Again I hope this is helpful - CC @andrius-senulis & @mjnitz02


c-h-russell-walker:
Closing in favor of:
https://github.com/datarobot/airflow-provider-datarobot/pull/111/

PR #112, opened 2025-01-21
Use 3.12 as primary dev version, support 3.9-3.13

tsdaemon:

## Summary

We should aim to use 3.12 for development, but clients of this package may use older versions of Python, therefore we should test and support of them. This:
* set's python 3.12 as primary version for CI and local development 
* updates requirement to allow versions 3.9 and older
* implements matrix to run tests on all supported versions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,17 +4,20 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "airflow-provider-datarobot"
-version = "0.0.10"
+version = "0.0.12"
 description = "DataRobot Airflow provider."
 readme = "README.md"
-requires-python = "~=3.11"
+requires-python = ">=3.9"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -4,17 +4,20 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "airflow-provider-datarobot"
-version = "0.0.10"
+version = "0.0.12"
 description = "DataRobot Airflow provider."
 readme = "README.md"
-requires-python = "~=3.11"
+requires-python = ">=3.9"
```
tsdaemon, line 10: @andrius-senulis here I overwrite your recent changes, hope its ok

andrius-senulis: Totally fine!

andrius-senulis (approved):

GunnerGlory (approved):

PR #114, opened 2025-01-21
[MLDEV-1143] Spike: Add initial version of automated operator generation code and associated tests

mjnitz02:

## Summary
This is not a 100% finished concept, but it does produce a working code parser. The code parser works as follows:
- A whitelist indicates the import module/package, classes within the package to generate operators for, and the unique methods that operators should be generated from.
- The code parses this whitelist and examines the docstrings of the appropriate methods.
- It uses the contents of the docstring to construct an operator from a basic template.
- The resulting combined code is then formatted using black and returned in a dictionary to the caller.

**This functionality will need to be evolved and expanded, but it provides the basic template for a quick method of producing operator python files in a 100% automated way.**

See `tests/fixtures/whitelist_test_output.txt` for an example result of this initial version.

## FUTURE/TODO
- This will be expanded to write/update the existing `python` files in the `operators` repo
  - **Currently this code is not generating any actual airflow operator files that can be used/tested. This will come in a followup.**
- This could be used in a pipeline to generate an automatically refreshed PR if changes are detected in existing operators (e.g. the python client changed the method signature, or improved the documentation)
- The `execute` functionality needs to be improved to provide a more generic interface/return that can be automatically parsed, but this is just an initial attempt at this concept.

## Changes
- Add the `autogen` package
- Add some test fixtures and a suite of tests that check the currently existing functionality against a single use case.
- Add initial `whitelist` which is extremely basic.


requirements.txt
```diff
@@ -14,3 +14,4 @@ pyenchant==3.2.2
 sphinx-copybutton
 sphinx-markdown-builder
 myst-parser==3.0.1
+black==24.10.0
```
mjnitz02, line 17: `ruff` doesn't provide a python native interface to format code structured as a string. This can allow us to construct automated code with an ok-level of accuracy initially (but with some formatting problems/inconsistencies) and then black can fix the actual formatting before returning it.

tests/fixtures/whitelist_test_output.txt
```diff
@@ -0,0 +1,46 @@
...
+    def execute(self, context: Context) -> str:
+        # Initialize DataRobot client
+        DataRobotHook(datarobot_conn_id=self.datarobot_conn_id).run()
+
+        if self.project_id is None:
+            raise ValueError("project_id is required for ModelJob.")
+        if self.model_job_id is None:
+            raise ValueError("model_job_id is required for ModelJob.")
+
+        result = datarobot.models.modeljob.ModelJob.get(
```
andrius-senulis, line 42: `datarobot` module is not imported in this code.

PR #115, opened 2025-01-22
[MLDEV-1170] Prepare script to generate stuff

tsdaemon:

## Summary

This:
1. Makes `generate_operator.py` into a script which can already be used to generate operators
2. Implements generation of imports
3. Adds POC generated operator to test the approach `datarobot_provider/operators/gen/modeljob/model_job_get_operator.py`

Problems:
Generated code gets minor fixes from `ruff`, I would prefer that we generate files which are always the same.
I have also added followups to https://datarobot.atlassian.net/browse/MLDEV-1169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,3 +80,6 @@ build-astro-dev:
 
 copy-examples-astro-dev:
 	cp -r ./datarobot_provider/example_dags/* ./astro-dev/dags/
+
+autogen-operators:
+	python ./datarobot_provider/autogen/generate_operator.py --whitelist ./datarobot_provider/autogen/whitelist.yaml --output_folder ./datarobot_provider/operators/gen
--- a/datarobot_provider/autogen/generate_operator.py
+++ b/datarobot_provider/autogen/generate_operator.py
@@ -45,11 +45,11 @@ class GenerateOperators:
 
         Parameters
         ----------
-        module_key
+        module_key : str
             The module name.
-        module_obj_key
+        module_obj_key : str
             The object name within the module.
-        method
+        method : str
             The method name.
         method_docstring : NumpyDocString
             The docstring for the method.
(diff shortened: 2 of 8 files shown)


tsdaemon:
@mjnitz02 1. Regarding real operators: I think we will need them to coexist with the generated in case we need something custom implemented. At least for the initial stage while we setting up the approach. I generally support the idea that we should implement custom logic as methods of public API client instead doing it here, but maybe we are yet missing something.
2. I was thinking about testing. Maybe its a good idea to generate tests along operators code. But on the other hand they are so simple, that its going to be just one test case.

tsdaemon:
@mjnitz02 also one thing that needs to be added. The currently implemented approach will work well for `classmethod`'s, but we also should support regular methods: e.g. `Project.get(project_id).start_modeling(...)`

demidov91:
Isn't `DecoratedOperator` doing the same? I mean, it converts a single python function with args/kwargs into an operator. So, we wouldn't have to generate any code.

mjnitz02:
@demidov91 I think one area we might have to do some custom handling would be figuring out how to return the response from everything in a object id is most all scenarios. Returning our actual "objects" wouldn't work as input to downstream operators. I think we could create a new base class though based off the `DecoratedOperator` and possibly leverage a lot of that there. We also want to make sure the code updated files have the correct docstrings and everything available for documentation, so we will still need some degree of automated parsing, even if we can help streamline the operator execute functionality.

mjnitz02 (approved):
I'm good with these changes. Thanks for continuing to move this concept forward! I think as we dig deeper into it we'll continue to improve it and maybe we can find some ways to make some portions simpler.

demidov91:
I don't block the PR but still convinced that code generation is always a bad idea.

tsdaemon:
@demidov91 can you share more, why do you think code generation is a bad idea? I believe its the best way we have to avoid writing and maintaining pretty mindless code: as operators proves to be simply wrappers around the python client.

Regarding your earlier remark about `DecoratedOperator`: we still need to write a function which will be decorated. So we can consider alternative generation strategy, where we generate methods with `DecoratedOperator` instead of classes: it will be simpler, but I wonder what are the downsides.

PR #116, opened 2025-01-23
[DM-16787] Add `CreateWranglingRecipeOperator` 

demidov91:

## Summary

Add a new operator: `CreateWranglingRecipeOperator`

## Rationale

Users should be able to create a recipe and use it in a recently added `CreateDatasetFromRecipeOperator`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,8 +3,9 @@
 ## Unreleased
 
 ### New features
-- Introduce `CreateDatasetFromRecipeOperator <datarobot_provider.operators.ai_catalog.CreateDatasetFromRecipeOperator>` 
-to publish a dataset based on existing wrangling recipe. 
+- Introduce `CreateWranglingRecipeOperator <datarobot_provider.operators.ai_catalog.CreateWranglingRecipeOperator>` 
+and `CreateDatasetFromRecipeOperator <datarobot_provider.operators.ai_catalog.CreateDatasetFromRecipeOperator>` 
+to create a wranglig recipe and publish it as a dataset into an existing experiment container. 
 
 ### Bugfixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ### New features
 - Introduce `CreateWranglingRecipeOperator <datarobot_provider.operators.ai_catalog.CreateWranglingRecipeOperator>` 
 and `CreateDatasetFromRecipeOperator <datarobot_provider.operators.ai_catalog.CreateDatasetFromRecipeOperator>` 
-to create a wranglig recipe and publish it as a dataset into an existing experiment container. 
+to create a wrangling recipe and publish it as a dataset into an existing use case.
 
 ### Bugfixes
 
--- a/datarobot_provider/operators/ai_catalog.py
+++ b/datarobot_provider/operators/ai_catalog.py
@@ -510,3 +510,53 @@ class CreateOrUpdateDataSourceOperator(BaseOperator):
             self.log.info(f"DataSource:{dataset_name} successfully created, id={data_source.id}")
 
         return data_source.id
+
+
+class CreateRecipeOperator(BaseOperator):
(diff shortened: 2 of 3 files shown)


datarobot_provider/operators/ai_catalog.py
```diff
@@ -444,10 +448,6 @@ class CreateOrUpdateDataSourceOperator(BaseOperator):
     :rtype: str
     """
 
-    # Specify the arguments that are allowed to parse with jinja templating
```
bdrosen96, line 447: why were these removed?

demidov91: Thanks. Actually, I wanted to remove the default fields in the I operator I've recently added, `CreateDatasetFromRecipeOperator`. Updated.

datarobot_provider/operators/ai_catalog.py
```diff
@@ -270,10 +273,6 @@ class CreateDatasetFromRecipeOperator(BaseOperator):
...
```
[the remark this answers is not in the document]
demidov91: Well, until it's explicitly required we can go on without it. It's out of the PR scope. Current fields are empty anyway.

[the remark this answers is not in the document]
bdrosen96: they are required in the constructor since there are no defaults.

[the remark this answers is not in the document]
demidov91: So, yes, user has to provide the values. They can do it in a python DAG.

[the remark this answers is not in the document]
bdrosen96: Why would we not have:

```
 template_fields: Sequence[str] = [
        "recipe_id",
]
```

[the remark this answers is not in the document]
demidov91: Added

datarobot_provider/operators/ai_catalog.py
```diff
@@ -266,13 +266,14 @@ class CreateDatasetFromRecipeOperator(BaseOperator):
...
-    :type materialization_schema: str
+    :type materialization_schema_param: str
     :param materialization_table_param: Name of the parameter in the configuration to use as materialization_table
-    :type materialization_table: str
+    :type materialization_table_param: str
     :return: DataRobot AI Catalog dataset ID
     :rtype: str
     """
 
+    template_fields = ["recipe_id"]
```
bdrosen96, line 276: Do we need do_snapshot as well? Or should we make that optional with a default in the constructor ?

demidov91: As far as I understand, user can set any value for any parameter regardless `template_fields` list. `template_fields` only makes it possible to use template variables in the definition. 

Users should be very clear about `do_snapshot` when defining the operation. We shouldn't define any default value as both options make sense.

bdrosen96 (approved):

CHANGES.md
```diff
@@ -3,8 +3,9 @@
 ## Unreleased
 
 ### New features
-- Introduce `CreateDatasetFromRecipeOperator <datarobot_provider.operators.ai_catalog.CreateDatasetFromRecipeOperator>` 
-to publish a dataset based on existing wrangling recipe. 
+- Introduce `CreateWranglingRecipeOperator <datarobot_provider.operators.ai_catalog.CreateWranglingRecipeOperator>` 
+and `CreateDatasetFromRecipeOperator <datarobot_provider.operators.ai_catalog.CreateDatasetFromRecipeOperator>` 
+to create a wranglig recipe and publish it as a dataset into an existing experiment container.
```
akshoop, line 8:
```suggestion
to create a wrangling recipe and publish it as a dataset into an existing experiment container/use case. 
```

akshoop (approved):

CHANGES.md
demidov91: Nice catch!

PR #117, opened 2025-01-23
[MLDEV-1178] Move existing dags to deprecated

mjnitz02:

## Summary
The plan is to remove these DAGs before the 11.0 release and our primary DAGs will be replaced with a set of P0 dags. Because there may be some useful reference material here, I am proposing we move all of the existing DAGs to a `deprecated` folder for now. I've also confirmed that not all of these DAGs actually work with the currently available versions of datarobot. We will add the new examples directly to the `example_dags` folder and when 11.0 comes around we'll delete this folder entirely.

## Rationale
- Cleanup examples in preparation for new P0 dags.


mariusvilkas (approved):
LGTM

c-h-russell-walker (approved):

PR #118, opened 2025-01-27
[DM-16785] Add `CreateFeatureDiscoveryRecipeOperator`

jtmcc:

## Summary
This is the first step to enabling Feature Discovery in workbench using AirFlow.
In this PR we add a way for users to create Feature Discovery recipes.

Subsequent PRs will add the following steps which will allow the full workflow to be performed:
- Add "recipe_id" parameter to `CreateProjectOperator` which will allow users to create and run projects from Feature Discovery recipes
- Create a `CreateDatasetFromRecipeOperator` which allows users to "publish" Feature Discovery datasets

### DIFF
--- a/datarobot_provider/operators/feature_discovery.py
+++ b/datarobot_provider/operators/feature_discovery.py
@@ -12,6 +12,7 @@ from typing import Optional
 
 import datarobot as dr
 from airflow.exceptions import AirflowException
+from airflow.exceptions import AirflowFailException
 from airflow.models import BaseOperator
 from airflow.utils.context import Context
 from datarobot import SNAPSHOT_POLICY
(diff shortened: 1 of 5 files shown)


datarobot_provider/operators/feature_discovery.py
```diff
@@ -21,6 +22,85 @@
...
+            "/recipes/fromDataset/", data={
+                'useCaseId': self.use_case_id,
+                'status': 'draft',
+                'datasetId': self.dataset_id,
+                'recipeType': 'FEATURE_DISCOVERY',
+            }
+        )
+        if response.status_code != 201:
+            e_msg = "Server unexpectedly returned status code {}"
+            raise AirflowFailException(e_msg.format(response.status_code))
```
jtmcc, line 85: This can be replaced by a Python Client call once a new version is released with this change:
https://github.com/datarobot/public_api_client/pull/3485

datarobot_provider/operators/feature_discovery.py
```diff
@@ -21,6 +22,85 @@
...
+    :type datarobot_conn_id: str, optional
+    :return: DataRobot Recipe ID
+    :rtype: str
+    """
+
+    # Specify the arguments that are allowed to parse with jinja templating
+    template_fields: Iterable[str] = [
+        "dataset_id",
+        "use_case_id",
+        "relationships_configuration_id",
```
jtmcc, line 45: There were two options we could choose to implement here:
- Users pass an ID of a Relationship Configuration they created and we **copy** the contents into the configuration that is created automatically below, OR
- Ask users to pass in the componenets of a relationship configuraiton (Dataset Defintions, Relationships, Settings) rather than a config to explicitly demonstrate that a new configuration will be created in the subsequent process

The former option is cleaner, but risks being slightly more confusing (a user might expect that changes made to their original Configuration will impact the recipe after being created, but that is not the case as we're using a duplicated configuration now). The latter option is clearer in the underlying process but more complex to implement.

I've opted for the simpler approach but am open to suggestions otherwise

demidov91: @artemoboznyi  I think we have to take the second approach: create a feature discovery recipe with all configurations in one step (one operation). It would make pipelines shorter. Additionally, I don't see an airflow scenario where users would re-use a relationships configuration id, i.e. work with relationships configuration independently.

artemoboznyi: agree, we should create a feature discovery recipe with all configurations in one step

jtmcc: Sure, I've updated to pass in the config components individually here instead

demidov91 (approved):

PR #119, opened 2025-01-27
[CFX-1398] Adding opt-in Airflow ruff checking

c-h-russell-walker:

## Summary

Proposal to also use Ruff's Airflow rules - docs here:
https://docs.astral.sh/ruff/rules/#airflow-air

## Rationale

I think since we're using Ruff in this repo. and there were minimal errors it's a good opportunity to enable these checks

## Updates

- Added Airflow ruff config so we check those rules
- Updated code to get rid of errors => `AIR001 Task variable name should match the `task_id``


mjnitz02 (approved):
I think this is a great change. I'm always 👍  for improving the linters. While they can be naggy, the keep stuff really consistent :D

PR #120, opened 2025-01-27
[CFX-1399] Adding notebook Operators, Sensors and an E2E DAG

c-h-russell-walker:

## Summary

Currently a DRAFT PR however I believe this can be close to production ready.

Adding Operators and Sensors for DR Notebooks and also making some needed code alterations like .gitignore for example.

## Rationale

This is the approach I'm most in favor of where we use granular steps/Operators in our notebook execution flow and leverage Airflow Sensors to handle the two needed polling states.

## TODOs

_Maybe_ parametrize the timeout used with the `NotebookExecutionCompleteSensor` in the example DAG since in DR scheduling we give an upper limit of 24 hours - in the code I've written so far I've set the timeout to 10 minutes - so for _sure_ we should make it longer than that by default.

## Screenshots

**Successful end-to-end run:**

Screenshot 2025-01-27 at 3 21 27 PM

<hr>

**Example passing notebook params in DAG:**

Screenshot 2025-01-27 at 3 42 31 PM

<hr>

**Example of Airflow Sensor checking/polling to see if notebook session is running:**

Screenshot 2025-01-27 at 4 43 33 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,9 +142,6 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
-# Jupyter notebooks
-notebooks/
-
 # Astro development environments
 astro-dev/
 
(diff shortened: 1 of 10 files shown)


c-h-russell-walker:
FYI - we're likely taking a different approach as seen here:

https://github.com/datarobot/airflow-provider-datarobot/pull/125/

PR #122, opened 2025-01-28
[DM-16904] Add "recipe_id" optional parameter to `CreateProjectOperator`

jtmcc:

## Summary
For UXR Feature Discovery we need to be able to create projects from Recipes. This PR updates to add an optional "recipe_id" to the create project operator.

### DIFF
--- a/datarobot_provider/operators/datarobot.py
+++ b/datarobot_provider/operators/datarobot.py
@@ -152,7 +152,7 @@ class CreateProjectOperator(BaseOperator):
             return project.id  # type: ignore[attr-defined, unused-ignore]
 
         elif self.recipe_id is not None:
-            response = dr.client.get_client().post("/projects/", json={"recipeId": self.recipe_id})
+            response = dr.client.get_client().post("/projects/", data={"recipeId": self.recipe_id})
 
             if response.status_code != 202:
                 e_msg = "Server unexpectedly returned status code {}"
--- a/datarobot_provider/operators/datarobot.py
+++ b/datarobot_provider/operators/datarobot.py
@@ -14,8 +14,8 @@ from airflow.exceptions import AirflowException
 from airflow.exceptions import AirflowFailException
 from airflow.models import BaseOperator
 from airflow.utils.context import Context
-
 from datarobot.utils.waiters import wait_for_async_resolution
+
 from datarobot_provider.hooks.datarobot import DataRobotHook
 
 DATAROBOT_MAX_WAIT = 3600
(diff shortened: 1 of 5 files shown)


datarobot_provider/operators/datarobot.py
```diff
@@ -99,9 +102,23 @@ def execute(self, context: Context) -> Optional[str]:
             )
             return project.id  # type: ignore[attr-defined, unused-ignore]
 
+        elif self.recipe_id is not None:
+            response = dr.client.get_client().post("/projects/", data={"recipeId": self.recipe_id})
```
jtmcc, line 106: Can be replaced by python client `Project.from_recipe` when the next version is released

datarobot_provider/operators/datarobot.py
```diff
@@ -99,9 +102,23 @@ def execute(self, context: Context) -> Optional[str]:
...
+
+            if response.status_code != 202:
+                e_msg = "Server unexpectedly returned status code {}"
+                raise AirflowFailException(e_msg.format(response.status_code))
+
+            project_id = response.json()["pid"]
```
bdrosen96, line 112: Do you need to have it poll for the location header location until it completes or errors? I believe python client does that for you when not using recipe id, so this should match that behavior

mariusvilkas (approved):
LGTM after addressing comments.

datarobot_provider/operators/datarobot.py
jtmcc: Good point, I've added in a call to wait for async resolution

bdrosen96: Should you assign project_id before calling wait_for_async_resolution ?

bdrosen96 (approved):

datarobot_provider/operators/datarobot.py
jtmcc: I've updated to use the python client instead

datarobot_provider/operators/datarobot.py
```diff
@@ -148,9 +151,19 @@ def execute(self, context: Context) -> Optional[str]:
             )
             return project.id  # type: ignore[attr-defined, unused-ignore]
 
+        elif self.recipe_id is not None:
+            project = dr.Project.create_from_recipe(
```
mjnitz02, line 155: @jtmcc Does this code actually work? It looks like `Project` doesn't have this function?

mjnitz02: I only see it under `dr.Dataset`?

mjnitz02: Ah wait nevermind, the datarobot was out of date :D

PR #123, opened 2025-01-28
[DM-16784] Add a GetDataStoreOperator

demidov91:

## Summary

* Get a `data_store_id` based on the connection name.
* Make `CreateOrUpdateDataSourceOperator` more flexible:
  * `dataset_name` parameter is now optional
  * All parameters are templates now.
* Create a wrangling recipe from a db table. No intermediate dataset needed. 

## Rationale

* Datarobot relies on default credentials heavily in UXR. New API endpoints don't expect plain credentials anymore. So, users have to **prefer Datarobot over Airflow connections management**.

* We're going to build an example DAG with database wrangling. The PR prepares existing operators for this.

### DIFF
--- a/datarobot_provider/operators/ai_catalog.py
+++ b/datarobot_provider/operators/ai_catalog.py
@@ -11,6 +11,7 @@ from collections.abc import Sequence
 from typing import Any
 from typing import List
 from typing import Optional
+from hashlib import sha256
 
 import datarobot as dr
 from airflow.exceptions import AirflowException
@@ -432,10 +433,20 @@ class CreateDatasetVersionOperator(BaseOperator):
 
 class CreateOrUpdateDataSourceOperator(BaseOperator):
     """
-    Creates the data source or updates it if its already exist and return data source ID.
+    Get an existing data source by name and update it if any of *table_schema*, *table_name*, *query* are specified.
+    Create a new data source or update a data source and return data source ID.
 
     :param data_store_id: DataRobot data store ID
     :type data_store_id: str
+    :param dataset_name: Data source canonical name to create or update.
+    :type dataset_name: Optional[str]
+    :param table_schema: Database schema name.
+    :type table_schema: Optional[str]
+    :param table_name: Database table name.
+    :type table_name: Optional[str]
+    :param query: Database table name.
+    :type query: Optional[str]
+
     :param datarobot_conn_id: Connection ID, defaults to `datarobot_default`
     :type datarobot_conn_id: str, optional
     :return: DataRobot AI Catalog data source ID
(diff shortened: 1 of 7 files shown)


datarobot_provider/operators/ai_catalog.py
```diff
@@ -431,32 +433,55 @@ def execute(self, context: Context) -> str:
...
     :rtype: str
     """
 
     # Specify the arguments that are allowed to parse with jinja templating
-    template_fields: Sequence[str] = ["data_store_id"]
-    template_fields_renderers: dict[str, str] = {}
-    template_ext: Sequence[str] = ()
+    template_fields: Sequence[str] = [
+        "data_store_id",
+        "dataset_name",
```
bdrosen96, line 459: Why do you use dataset_name here? Wouldn't datasource_name be better?

datarobot_provider/operators/ai_catalog.py
```diff
@@ -470,44 +495,67 @@ def execute(self, context: Context) -> Optional[str]:
...
+            self.log.info(
+                f"DataSource:{self.dataset_name} successfully created, id={data_source.id}"
             )
-            self.log.info(f"DataSource:{dataset_name} successfully created, id={data_source.id}")
 
         return data_source.id
 
+    def _get_default_data_source_name(self, data_store_id) -> str:
+        """Build default name based on the data source params."""
+        parts = ["Airflow", data_store_id]
```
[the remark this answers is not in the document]
demidov91: Let's say, this is `public.transactions` table. I want the default name to be `Airflow-<data_store_id>-table-pubic-transactions`. So, the next time we launch a pipeline it uses an existing data source instead of creating a new one with a unique name.

datarobot_provider/operators/ai_catalog.py
demidov91: That's for 2 reasons. I don't want to break existing pipelines, whoever uses the operator. Also, the naming makes some sense. Users could set `dataset_name` once in their context and use it both for a dataset and datasource name.

datarobot_provider/operators/ai_catalog.py
```diff
@@ -470,44 +495,67 @@ def execute(self, context: Context) -> Optional[str]:
         data_store = dr.DataStore.get(data_store_id=self.data_store_id)
         self.log.debug(f"Found existing DataStore: {data_store.canonical_name}, id={data_store.id}")
 
-        dataset_name = context["params"]["dataset_name"]
```
bdrosen96, line 473: How will it use the context now? Does the contructor templaing with params automatically refer to context params?

demidov91: All `template_fields` get their values according to the templates if user hasn't defined anything else explicitly. So, for `dataset_name` the template is `{{ params.get('dataset_name', '') }}` which is effectively `context["params"].get("dataset_name", "")`

bdrosen96 (approved):

tsdaemon (commented):
Please add add a note to CHANGES

demidov91:
done

tsdaemon (approved):
LGTM

PR #124, opened 2025-01-28
[MLDEV-1189] Add very basic hospital readmissions dag that we can build from

mjnitz02:

## Summary
This is a very simple dag, but it does run correctly and pretty fast on local or staging. All you need is to define a basic `datarobot_connection` and it runs.

## Rationale
Dag that does the following:
- Uploads a dataset (We can replace this with the wrangler pipeline)
- Creates a project
- Starts a Project
- Checks for autopilot completion

There are additional stages we need to add for the P0, but this gets us a running template to build from.


mariusvilkas (approved):
Not very familiar with this code. It looks ok to me.
LGTM

PR #126, opened 2025-01-30
[MLDEV-1174] Add use case create operator

mjnitz02:

## Summary
This was one of the remaining MLDEV operators blocking the P0 DAG. This allows a pipeline to create a use case. It's a pretty simple operator. Also the docstring has been moved to numpy now so that it works with the docs properly.

## Rationale
- Add a new create use case operator
- Add a simple test

Followup:
We probably need to investigate if we should have a more robust operator/dag test strategy in place. It feels like the existing flow heavily just relies on mocking out the core pySDK call, and this doesn't necessarily check that the upstream call hasn't changed in any way.

### DIFF
--- a/datarobot_provider/operators/datarobot.py
+++ b/datarobot_provider/operators/datarobot.py
@@ -24,7 +24,7 @@ DATETIME_FORMAT = "%Y-%m-%d %H:%M:%s"
 
 class CreateUseCaseOperator(BaseOperator):
     """
-    Creates DataRobot project.
+    Creates a DataRobot Use Case.
 
     Parameters
     ----------
@@ -33,11 +33,11 @@ class CreateUseCaseOperator(BaseOperator):
 
     Returns
     -------
-    str: DataRobot project ID
+    str: DataRobot UseCase ID
     """
 
     # Specify the arguments that are allowed to parse with jinja templating
-    template_fields: Sequence[str] = ["dataset_id", "dataset_version_id", "credential_id"]
+    template_fields: Sequence[str] = ["credential_id"]
     template_fields_renderers: dict[str, str] = {}
     template_ext: Sequence[str] = ()
     ui_color = "#f4a460"
--- a/datarobot_provider/operators/datarobot.py
+++ b/datarobot_provider/operators/datarobot.py
@@ -22,6 +22,55 @@ DATAROBOT_AUTOPILOT_TIMEOUT = 86400
 DATETIME_FORMAT = "%Y-%m-%d %H:%M:%s"
 
 
+class CreateUseCaseOperator(BaseOperator):
+    """
+    Creates DataRobot project.
+
+    Parameters
+    ----------
+    datarobot_conn_id: str
+        Connection ID, defaults to `datarobot_default`
+
+    Returns
+    -------
+    str: DataRobot project ID
+    """
+
+    # Specify the arguments that are allowed to parse with jinja templating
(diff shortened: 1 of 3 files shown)


datarobot_provider/operators/datarobot.py
```diff
@@ -22,6 +22,55 @@
 DATETIME_FORMAT = "%Y-%m-%d %H:%M:%s"
 
 
+class CreateUseCaseOperator(BaseOperator):
+    """
+    Creates DataRobot project.
```
mbromanowsky, line 27: Nope: a Use Case can contain multiple Experiments aka Projects. (Looks like this might be copypasta.)

```suggestion
    Creates a DataRobot Use Case.
```

datarobot_provider/operators/datarobot.py
```diff
@@ -22,6 +22,55 @@
...
+
+    Parameters
+    ----------
+    datarobot_conn_id: str
+        Connection ID, defaults to `datarobot_default`
+
+    Returns
+    -------
+    str: DataRobot project ID
```
mbromanowsky, line 36:
```suggestion
    str: DataRobot UseCase ID
```

datarobot_provider/operators/datarobot.py
```diff
@@ -22,6 +22,55 @@
...
+    """
+
+    # Specify the arguments that are allowed to parse with jinja templating
+    template_fields: Sequence[str] = ["dataset_id", "dataset_version_id", "credential_id"]
```
mbromanowsky, line 40: Are these copied over from the `CreateProjectOperator`? Don't you instead need `use_case_name` and `use_case_description` here?

mbromanowsky (commented):
Seems like a few small changes are needed, especially with the templated arguments. Or am I just confused?

datarobot_provider/operators/datarobot.py
mjnitz02: It is copied, but actually most of the values are passed in the `x_com context` which is defined manually and optionally by the user in the airflow UI. Because these aren't required to actually be there, I allowed them through `x_com` only, so they aren't listed in the `template_fields`. I need to remove the two ids though!

mjnitz02, line 40:
```suggestion
    template_fields: Sequence[str] = ["credential_id"]
```

docs/autodoc/operators_datarobot.md
```diff
@@ -0,0 +1,8 @@
+(datarobot-modules)=
```
mbromanowsky, line 1: It's great to see docs added! Are there any tests to build the docs? Or at least instructions in the README for building the docs? Not required for this PR, just a note for the future if needed.

mbromanowsky (approved):
LGTM for MLDev

docs/autodoc/operators_datarobot.md
c-h-russell-walker: Sorry I'm creeping PRs in this repo.

Great question - I'm hopefully going to have a Notebooks PR ready at some point and would like to make sure I have docs included as well.

mjnitz02: @mbromanowsky @c-h-russell-walker In the docs folder there is a `make test-docs` but its not running in a formal pipeline yet. We still need to get this pipeline added to harness and all the PRs

mjnitz02: We have some tickets related to this here: https://datarobot.atlassian.net/browse/MLDEV-1131

PR #129, opened 2025-02-03
[EXPERIMENTAL] DatarobotBaseOperator

demidov91:

## Summary

Introduce a base class for all Datarobot operators. Apply it to operators modules from A to F. More updates to do in a subsequent PR.

## Rationale

There is quite a lot of repeatable codebase in the repo. Move it into the base class.

### DIFF
--- a/datarobot_provider/operators/base_datarobot_operator.py
+++ b/datarobot_provider/operators/base_datarobot_operator.py
@@ -6,7 +6,8 @@
 #
 # Released under the terms of DataRobot Tool and Utility Agreement.
 
-from typing import Optional, Any
+from typing import Any
+from typing import Optional
 
 import datarobot as dr
 from airflow.exceptions import AirflowException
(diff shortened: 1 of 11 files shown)


datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class BaseDatarobotOperator(BaseOperator):
+    ui_color = "#f4a460"
+
+    dr_hook: DataRobotHook
+    min_version: Optional[str] = None
+    requires_early_access = False
+
+    def pre_execute(self, context: Any):
```
andrius-senulis, line 27: Is this built in method in Airflow that will be called automatically before `execute()`?

demidov91: Correct

datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -0,0 +1,65 @@
...
+        super().pre_execute(context)
+
+        self.check_dr_client_version()
+        self.dr_hook = DataRobotHook(datarobot_conn_id=self.datarobot_conn_id)
+        self.dr_hook.run()
+
+        self.validate()
+
+    def check_dr_client_version(self):
```
c-h-russell-walker, line 36: First: I _really_ like what you've done in general here and think totally a valid base class to add 👍 

I _do_ know that there's on-going (needed to unblock DR Notebooks work) discussion related to how we're trying/going to handle DR Python Client version(s).

I'd argue that _if_ we're not going to leverage the `datarobot-early-access` library then we shouldn't check for it here.

Essentially I think this PR is great but maybe as far as implementing (or not?) a `check_dr_client_version()` we should wait on whatever deployment strategy decision we end up making.
CC @asuslov to see what yout hink here.

demidov91: I agree, we shouldn't freeze in the code what's not decided on the product side. It's more like a demo of what (and how easy!) can be done. I'll remove the method in the final PR.

datarobot_provider/operators/datarobot.py
```diff
@@ -38,29 +36,18 @@ class CreateUseCaseOperator(BaseOperator):
...
     template_fields: Sequence[str] = ["credential_id"]
-    template_fields_renderers: dict[str, str] = {}
-    template_ext: Sequence[str] = ()
-    ui_color = "#f4a460"
 
     def __init__(
         self,
         *,
         credential_id: Optional[str] = None,
-        datarobot_conn_id: str = "datarobot_default",
```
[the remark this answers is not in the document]
demidov91: We can (and should) define it in the docstring while it's a an argument of the base class.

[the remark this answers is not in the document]
mjnitz02: Does this cause a linter issue if it doesn't see it? I could be totally wrong, for some reason I thought it raised a warning if you add things to the docstrings that are missing. This might have been some local plugin I had enabled though

PR #130, opened 2025-02-03
[MLDEV-1194] Create _experimental folder and move untested dags to this

mjnitz02:

## Summary
After discussions today the decision is to go ahead with a `airflow-provider-datarobot` and a `airflow-provider-datarobot-early-access` package. Anything that is not in scope of 11.0 or planned for formal QA should be moved over to the `_experimental` folder. This won't affect the testing as we just run all the tests for everything. This will primarily affect which datarobot package is installed on the backend.

## Rationale
Allow early-access initial concept.


datarobot_provider/_experimental/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+import logging
+
+logger = logging.getLogger(__package__)
+
+experimental_warning = (
+    "You have imported from the _experimental directory.\n"
+    "This directory is used for unreleased datarobot features.\n"
+    "Unless you specifically know better,"
+    " you don't have the access to use this functionality in the app, so this code will not work."
+)
```
mbromanowsky, line 17: This doesn't read quite right to me, because my assumption is that "access" isn't the issue -- that is, feature flags will not be the reason for things not working. Maybe:

```suggestion
experimental_warning = (
    "You have imported from the _experimental directory.\n"
    "This directory is used for unreleased datarobot features.\n"
    "Code from this directory may be untested and may change without warning."
)
```

mbromanowsky (approved):
LGTM for MLDev, with one suggestion about the warning message

datarobot_provider/_experimental/__init__.py
c-h-russell-walker: We could call in the Docs. reinforcements as well 😄 

@nate-goudreault if you feel like chiming in here.

nate-goudreault:
```suggestion
experimental_warning = (
    "You have imported from the _experimental directory.\n"
    "This directory is used for unreleased DataRobot features.\n"
    "Unless you specifically know better,"
    " you don't have the access to use this functionality in the app, so this code will not work."
)
```

PR #131, opened 2025-02-04
[DM-16986] Introduce a `BaseDatarobotOperator`

demidov91:

## Summary

Introduce a base class for all Datarobot operators. Apply it to operators modules from `a...` to `f...`. More updates to do in a subsequent PR.

## Rationale

There is quite a lot of repeatable codebase in the repo. Move it into the base class.

### DIFF
--- a/datarobot_provider/operators/base_datarobot_operator.py
+++ b/datarobot_provider/operators/base_datarobot_operator.py
@@ -6,7 +6,8 @@
 #
 # Released under the terms of DataRobot Tool and Utility Agreement.
 
-from typing import Optional, Any
+from typing import Any
+from typing import Optional
 
 import datarobot as dr
 from airflow.exceptions import AirflowException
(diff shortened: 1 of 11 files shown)


jtmcc (approved):
Nice. Looks good

datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class BaseDatarobotOperator(BaseOperator):
+    ui_color = "#f4a460"
+
+    dr_hook: DataRobotHook
+    min_version: Optional[str] = None
+    requires_early_access = False
+
+    def pre_execute(self, context: Any):
```
c-h-russell-walker, line 25: Is it possible to annotate `context` here with maybe:
`from airflow.utils.context import Context`

That is what we annotate `execute(...` with

c-h-russell-walker (approved):
I think this is a great improvement to the codebase 👍 
Thanks for doing this.

datarobot_provider/operators/base_datarobot_operator.py
demidov91: Added

mjnitz02 (approved):
Thanks for doing this, I think it really simplifies the overall operators and gets some of the duplicate hooking code out everywhere. 👍

PR #133, opened 2025-02-04
[MLDEV-1208] Add triggers and input set yaml updates to allow running commits on main

mjnitz02:

### DIFF
--- a/.harness/ci_checks_main_trigger.yaml
+++ b/.harness/ci_checks_main_trigger.yaml
@@ -1,0 +1,28 @@
+trigger:
+  name: ci_checks_merge_trigger
+  identifier: ci_checks_merge_trigger
+  enabled: true
+  stagesToExecute: []
+  description: ""
+  tags: {}
+  encryptedWebhookSecretIdentifier: ""
+  orgIdentifier: AGENTS
+  projectIdentifier: airflowproviderdatarobot
+  pipelineIdentifier: cichecks
+  source:
+    type: Webhook
+    spec:
+      type: Github
+      spec:
+        type: Push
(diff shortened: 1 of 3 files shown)


.harness/ci_checks_pr_trigger.yaml
```diff
@@ -0,0 +1,31 @@
+trigger:
```
mjnitz02, line 1: This trigger is already active, but doesn't have a yaml stored in github.

.harness/ci_checks_main_trigger.yaml
```diff
@@ -0,0 +1,28 @@
+trigger:
```
mjnitz02, line 1: This is a new trigger to run checks on https://github.com/datarobot/airflow-provider-datarobot/commits/main/. This way we will get the green check or red X on each release. This can help us figure out if we have a merge conflict quickly in the repo and where/when it occurred. The overhead for these tests is very low and they are pretty fast.

c-h-russell-walker (approved):
Smart. 💯

.harness/ci_checks_pr_trigger.yaml
```diff
@@ -0,0 +1,31 @@
...
+  pipelineIdentifier: cichecks
+  source:
+    type: Webhook
+    spec:
+      type: Github
+      spec:
+        type: PullRequest
+        spec:
+          connectorRef: account.svc_harness_git1
+          autoAbortPreviousExecutions: true
```
mjnitz02, line 20: I added this so on PR updates we'll abort the previous pipeline executions. Small change to save on run resources

PR #134, opened 2025-02-04
[DM-16908] Better dag

demidov91:

## Summary

* Use csv upload in the example DAG
* Reuse a use case by name
* Add more operations into the example recipe

## Rationale
Provide as smooth quick start experience as possible:
* Use public data with no additional configurations needed
* Create and reuse a use case

### DIFF
--- a/datarobot_provider/operators/ai_catalog.py
+++ b/datarobot_provider/operators/ai_catalog.py
@@ -16,6 +16,7 @@ from typing import Optional
 import datarobot as dr
 from airflow.exceptions import AirflowException
 from airflow.utils.context import Context
+from datarobot.models.recipe_operation import RandomSamplingOperation
 
 from datarobot_provider.hooks.connections import JDBCDataSourceHook
 from datarobot_provider.operators.base_datarobot_operator import BaseDatarobotOperator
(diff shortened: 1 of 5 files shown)


datarobot_provider/operators/ai_catalog.py
```diff
@@ -29,43 +30,58 @@ class UploadDatasetOperator(BaseDatarobotOperator):
     Uploading local file to DataRobot AI Catalog and return Dataset ID.
     :param file_path: The path to the file.
     :type file_path: str, optional
-    :param file_path_param: Name of the parameter in the configuration to use as file_path, defaults to `dataset_file_path`
+    :param file_path_param: DEPRECATED. Name of the parameter in the configuration to use as file_path, defaults to `dataset_file_path`
```
bdrosen96, line 33: why is this deprecated?

datarobot_provider/operators/ai_catalog.py
```diff
@@ -29,43 +30,58 @@ class UploadDatasetOperator(BaseDatarobotOperator):
...
         # Upload Dataset to AI Catalog
         self.log.info("Upload Dataset to AI Catalog")
-        if self.file_path is None:
+        if not self.file_path and self.file_path_param:
+            self._file_path_param_is_deprecated()
             self.file_path = context["params"][self.file_path_param]
 
-        ai_catalog_dataset: dr.Dataset = dr.Dataset.create_from_file(
-            file_path=self.file_path,
-            max_wait=DATAROBOT_MAX_WAIT_SEC,
```
bdrosen96, line 63: This appears to have reduced the wait time considerably. Do we really want that?

datarobot_provider/operators/ai_catalog.py
demidov91: Users can pass `file_path={{ params.custom_param_name }}` instead of `file_path_param="custom_param_name"`. That reduces number of the operator parameters.

datarobot_provider/operators/ai_catalog.py
demidov91: Updated

bdrosen96 (approved):

mjnitz02 (approved):
I think these changes look good. Next week we can come back and do a passover of the strings to make sure the docs are pulling properly everywhere.

PR #137, opened 2025-02-05
[EXPERIMENTAL] Add a base class for a generic python function operator.

demidov91:

## Summary

Declare very short operator classes instead of generating them. Here is an example of the class autodoc:

Screenshot from 2025-02-05 17-35-22


mjnitz02:
@demidov91 Should we close this?

demidov91:
We can close it. Just keep the branch in case if we would like to revisit it.

demidov91:
Well, I did it myself :)

PR #140, opened 2025-02-06
[MLDEV-1207] New early access test release pipeline

mjnitz02:

## Summary
This adds a new early access release pipeline with test as the default option here. Once this works ok, I'll create a separate mirror with `pypi` as the `BUILD_TYPE` which should formally release early access to pypi.

See the test build on test.pypi for the result of the pipeline running in test mode

## TestPyPi Build
https://test.pypi.org/project/airflow-provider-datarobot-early-access/

### DIFF
--- a/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
+++ b/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
@@ -11,9 +11,10 @@ template:
     spec:
       cloneCodebase: true
       caching:
-        enabled: true
+        enabled: false
+        paths: []
       buildIntelligence:
-        enabled: true
+        enabled: false
       infrastructure:
         type: KubernetesDirect
         spec:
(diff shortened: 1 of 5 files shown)


setup_early_access.py
```diff
@@ -20,7 +20,7 @@
 version += datetime.today().strftime(".%Y.%m.%d")
 
 classifiers = [
-    "Development Status :: 4 - Preview",
+    "Development Status :: 4 - Beta",
```
mjnitz02, line 23: This has to be defined from the list here https://pypi.org/classifiers/

.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
```diff
@@ -11,9 +11,10 @@ template:
     spec:
       cloneCodebase: true
       caching:
```
mjnitz02, line 13: Some tweaks needed to this file to properly connect to twine and secrets correctly

c-h-russell-walker (approved):
Nice, LGTM 🚀

PR #144, opened 2025-02-07
[DM-17022] Default use case

demidov91:

## Summary

* Introduce a notion of a **default Use Case** for a DAG.
* Add a `BaseUseCaseEntityOperator` class
* Rename `CreateUseCaseOperator` into `GetOrCreateUseCaseOperator`

## Rationale

There are quite many DataRobot assets which have to belong to a UseCase. Define a default one in the very beginning of your DAG to avoid repetitive `use_case_id` parameters.

### DIFF
--- a/datarobot_provider/operators/base_datarobot_operator.py
+++ b/datarobot_provider/operators/base_datarobot_operator.py
@@ -71,7 +71,7 @@ class BaseUseCaseEntityOperator(BaseDatarobotOperator):
     def get_use_case_id(self, context: Context, required=False) -> Optional[str]:
         """Get self.use_case_id or a default Use Case id defined at runtime.
         Raises an exception if no Use Case id is defined and required=True"""
-        if use_case_id := (self.use_case_id or self.xcom_pull(context, XCOM_DEFAULT_USE_CASE_ID)):
+        if use_case_id := (self.use_case_id or self.xcom_pull(context, key=XCOM_DEFAULT_USE_CASE_ID)):
             return use_case_id
 
         if required:
--- a/datarobot_provider/operators/base_datarobot_operator.py
+++ b/datarobot_provider/operators/base_datarobot_operator.py
@@ -46,17 +46,26 @@ class BaseDatarobotOperator(BaseOperator):
 
 
 class BaseUseCaseEntityOperator(BaseDatarobotOperator):
+    """Base class for all operators which create a Use Case assets,
+    i.e. dataset, recipe, project, deployment, notebook, application, etc.
+
+    It introduces *use_case_id* template parameter and helper methods
+    to use a default Use Case if no Use Case was defined in the operator itself or a DAG context.
+    """
+
+    @classmethod
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if "use_case_id" not in cls.template_fields:
+            cls.template_fields = (*cls.template_fields, "use_case_id")
+            cls.logger().info(
+                "use_case_id is implicitly added into %s template_fields list.", cls.__name__
+            )
+
     def __init__(
         self, *, use_case_id: Optional[str] = "{{ params.use_case_id | default('') }}", **kwargs
     ):
         super().__init__(**kwargs)
-
-        if "use_case_id" not in self.template_fields:
-            raise AirflowException(
-                f"You must add use_case_id into {self.__class__.__name__} template_fields list "
-                "in order to use BaseUseCaseEntityOperator"
-            )
-
         self.use_case_id = use_case_id
 
     def get_use_case(self, context: Context, required=False) -> Optional[dr.UseCase]:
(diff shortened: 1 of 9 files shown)


datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -41,3 +43,70 @@ def __init__(self, *, datarobot_conn_id: str = "datarobot_default", **kwargs: An
...
+    ):
+        """Add an *entity* into Use Case defined as self.use_case_id or a default DAG Use Case."""
+        if use_case := self.get_use_case(context):
+            use_case.add(entity)
+            self.log.info(
+                '%s is added into use case "%s"', entity.__class__.__name__, use_case.name
+            )
+
+        else:
+            self.log.info("The %s won't be added into any use case.", entity.__class__.__name__)
```
demidov91, line 112: @c-h-russell-walker notebooks may find this base class useful.

c-h-russell-walker: Thank you!
Also I appreciate your review/comments on my draft PR - I've been sidelined by Fedramp 10.2 issues though so haven't gotten a chance to fully review/respond yet.

bdrosen96 (approved):

datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -41,3 +43,70 @@ def __init__(self, *, datarobot_conn_id: str = "datarobot_default", **kwargs: An
...
+    def get_use_case(self, context: Context, required=False) -> Optional[dr.UseCase]:
+        """Get a `dr.UseCase` entity based on self.use_case_id or a default Use Case defined at runtime.
+        Raises an exception if no Use Case is defined and required=True
+        """
+        if use_case_id := self.get_use_case_id(context, required=required):
+            return dr.UseCase.get(use_case_id)
+
+        return None
+
+    def get_use_case_id(self, context: Context, required=False) -> Optional[str]:
```
mjnitz02, line 80: Should required be a property of the Operator itself?

datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -41,3 +43,70 @@ def __init__(self, *, datarobot_conn_id: str = "datarobot_default", **kwargs: An
...
+
+        return None
+
+    def add_into_use_case(
+        self,
+        entity: Union[dr.Project, dr.Dataset, dr.models.Recipe, dr.Application],
+        *,
+        context: Context,
+    ):
+        """Add an *entity* into Use Case defined as self.use_case_id or a default DAG Use Case."""
```
mjnitz02, line 104: Is this mainly for adding datasets to a use case? I guess these are defined separately and then added to a use case? Sorry this might just be my misunderstanding. I guess like, a project, has to have a use case assigned to it during creation for it to be within the context of a use case?

datarobot_provider/operators/base_datarobot_operator.py
```diff
@@ -41,3 +43,70 @@ def __init__(self, *, datarobot_conn_id: str = "datarobot_default", **kwargs: An
...
+            cls.template_fields = (*cls.template_fields, "use_case_id")
+            cls.logger().info(
+                "use_case_id is implicitly added into %s template_fields list.", cls.__name__
+            )
+
+    def __init__(
+        self, *, use_case_id: Optional[str] = "{{ params.use_case_id | default('') }}", **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.use_case_id = use_case_id
```
mjnitz02, line 69: I was thinking maybe this could auto-initialize, but after looking I guess `context` only exists in the scope of `execute`, so it wouldn't be present. It might be worth considering as some followup if we could sort of autowrap this so that the methods just use `use_case_id = self.use_case_id` and its either populated or `None`. I'm not 100% sure if thats actually possible to do in the context of how airflow works though. Just wanted to put the idea out there.

mjnitz02 (approved):
I think this looks good. A few comments and questions as I'm trying to understand everything here.

datarobot_provider/operators/base_datarobot_operator.py
mjnitz02: It seems like based on the operator type, required would be pre-determined, or at least defined through the `init`?

datarobot_provider/operators/base_datarobot_operator.py
demidov91: It depends on the operator implantation. Yes, at the moment we use it for datasets. Nevertheless, it's applicable for projects or deployments, I guess

datarobot_provider/operators/base_datarobot_operator.py
demidov91: There are 3 ways to define a use case: an explicit parameter, a context variable and a default use case defined via xcom.

`__init__` can handle only the first option. 

Mandatory vs optional use_case_id could be really defined using `requires_use_case: bool` class field. At the moment PR uses an explicit argument in  `get_use_case_id()` method. I'm ok with either option. I could switch to the class field either in this PR or in a followup if you find that option better.

datarobot_provider/operators/base_datarobot_operator.py
demidov91: `__init__` is not context-aware by design because operators are initialized much before the execution.

PR #146, opened 2025-02-11
[MLDEV-1220] Fix locations of deployment operators (refactor)

mjnitz02:

## Summary
Going through the tasks under Build AI, and I noticed the deployment operators were split across two areas. I want to move the `datarobot.py` file to be more of a `project/use case` type file for basic functions. There is already an existing `deployment.py` set of operators, so this just combines all the existing deployment operators under that file.

### NONE OF THE OPERATORS ARE RE-WRITTEN HERE, THIS IS JUST REFACTORING THE EXISTING CODE


c-h-russell-walker (approved):
I think (IMHO) this is a great update - makes sense organizationally to have a smaller `datarobot.py` and split out operators to more "namespaced" files where/when appropriate - and added bonus of more granular CODEWONERS possible that way as well 👍

mbromanowsky (approved):
LGTM for MLDEV

eugene-stepanov-dr:
@mjnitz02 Is this refactoring backward-compatible with the library's public interface?

mjnitz02:
@eugene-stepanov-dr Right now this repo isn't really considered a `production` grade facing package. For 11.0 the goal is to make this a GA level repo and have everything in place for that. We're also deprecating and moving around any unused functions, or things that won't be formally QA tested.

The current state of the repo releases is that none of the existing operators have ever been QA'd or signed off and the package also has many operators which either don't work, or don't work correctly. So up until the 11.0 release we can basically move around, refactor and fix everything as if its a alpha/public preview set of code. After 11.0 the repo will have to ensure backwards compatibility from then on.

eugene-stepanov-dr (approved):
@mjnitz02 Thank you for the detailed explanation! It looks good to proceed further.

PR #147, opened 2025-02-11
[MLDEV-1165] Remove requirements.txt from repo

mjnitz02:

## Summary
This has been replaced with `setup` and `setuptools` variations everywhere. These allow for the different installations such as `pip install ....[dev]` or `[docs]` for running and building those when needed and keeps the base requirements very light.

At this point, all of the active testing and deployment pipeline should be migrated to using the `Makefile` commands, which all use the `setuptools` requirements, so the `requirements.txt` is just a legacy holdover and can be removed now.

-----------------
Just for full transparency, there is one pipeline (.harness/build_and_publish_pipy.yaml) which still technically references the `requirements.txt`, but this is the legacy pypi release pipeline and has been replaced by `.harness/release_pypi.yaml` which uses a common backend that is shared with the `early-access` pipelines and supports additional publications to `test.pypi`. This pipeline needs to be removed from the repo and our harness CI/CD in a followup. The pipeline doesn't actively run either, so nothing is at risk of breaking.


mbromanowsky (approved):
LGTM for MLDEV

c-h-russell-walker (approved):
Thank you for doing this.

PR #149, opened 2025-02-11
[DM-17021] Get catalogVersionId if not defined by users

jtmcc:

## Summary
In the normal Feature Discovery workflow we require version ID to be explicitly defined by the user.
We can simplify the user workflow for Airflow if we automatically get the version ID associated with a dataset when it isn't defined explicitly by the user.

### DIFF
--- a/datarobot_provider/operators/feature_discovery.py
+++ b/datarobot_provider/operators/feature_discovery.py
@@ -86,7 +86,7 @@ class CreateFeatureDiscoveryRecipeOperator(BaseUseCaseEntityOperator):
         # Get dataset version ID if it isn't defined by the user:
         for dataset_definition in self.dataset_definitions:
             if not dataset_definition.get("catalogVersionId"):
-                dataset = dr.Dataset.get(dataset_definition['catalogId'])
+                dataset = dr.Dataset.get(dataset_definition["catalogId"])
                 dataset_definition["catalogVersionId"] = dataset.version_id
 
         recipe = response.json()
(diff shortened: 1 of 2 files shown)


datarobot_provider/operators/feature_discovery.py
```diff
@@ -83,6 +83,12 @@ def execute(self, context: Context) -> None:
             e_msg = "Server unexpectedly returned status code {}"
             raise AirflowFailException(e_msg.format(response.status_code))
 
+        # Get dataset version ID if it isn't defined by the user:
+        for dataset_definition in self.dataset_definitions:
+            if not dataset_definition.get("catalogVersionId"):
+                dataset = dr.Dataset.get(dataset_definition["catalogId"])
+                dataset_definition["catalogVersionId"] = dataset.version_id
```
bdrosen96, line 90: Should we care about what snapshot policy is? Would we want to leave version id blank for certain snapshot policies?

jtmcc: Version ID should actually never be allowed to be blank. In DataRobot we enforce that both Catalog and Catalog Version IDs must be specified by the user at all times in Feature Discovery:
https://github.com/datarobot/DataRobot/blob/38cf51028b59def6f63bd3381c205e271aebe4fb/MLData/feature_discovery/public_api/relationships_config/validators.py#L137-L142

For Workbench integration in AirFlow, there isn't a simple way for the user to get Version ID after creating a dataset via say `UploadDatasetOperator`. After discussing with @demidov91 we're considering automatically grabbing version ID if a user doesn't define it specifically in AirFlow, but it would be up to the user to make sure the snapshot policy is aligned with the correct version if they choose not to define the version.

bdrosen96 (approved):

mjnitz02 (approved):
👍

PR #150, opened 2025-02-12
[MLDEV-1102] Blueprint select operator and some refactors

mjnitz02:

## Summary
This adds a select blueprint operator as well as fixing up some other stuff.

## Changes
- Renamed `model_training` to `model` to coincide with the `pySDK` better
- Change `TrainModelsOperator` to `AnalyzeAndModelProjectOperator` and change `set_target` command to use `analyze_and_model`
- Add `GetProjectBlueprintsOperator`
- Add tests
- Fix existing tests
- Fix docs for `datarobot operators`

### DIFF
--- a/datarobot_provider/_experimental/example_dags/model_retrain_dag.py
+++ b/datarobot_provider/_experimental/example_dags/model_retrain_dag.py
@@ -12,7 +12,7 @@ from airflow.decorators import dag
 from airflow.decorators import task
 from datarobot import SCORING_TYPE
 
-from datarobot_provider.operators.model import RetrainModelOperator
+from datarobot_provider.operators.model_training import RetrainModelOperator
 from datarobot_provider.sensors.model_training import ModelTrainingJobSensor
 
 
--- a/datarobot_provider/_experimental/example_dags/model_retrain_dag.py
+++ b/datarobot_provider/_experimental/example_dags/model_retrain_dag.py
@@ -12,7 +12,7 @@ from airflow.decorators import dag
 from airflow.decorators import task
 from datarobot import SCORING_TYPE
 
-from datarobot_provider.operators.model_training import RetrainModelOperator
+from datarobot_provider.operators.model import RetrainModelOperator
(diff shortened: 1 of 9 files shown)


datarobot_provider/operators/datarobot.py
```diff
@@ -266,4 +266,34 @@ def execute(self, context: Context) -> None:
                 f"Starting DataRobot Autopilot for project_id={project.id} "
                 f"with settings={context['params']['autopilot_settings']}"
             )
-            project.set_target(**context["params"]["autopilot_settings"])
+            project.analyze_and_model(**context["params"]["autopilot_settings"])
```
mjnitz02, line 269: The `set_target` functionality is deprecated and slated for removal. This needs to be `analyze_and_model`

datarobot_provider/_experimental/example_dags/model_retrain_dag.py
```diff
@@ -12,7 +12,7 @@
 from airflow.decorators import task
 from datarobot import SCORING_TYPE
 
-from datarobot_provider.operators.model_training import RetrainModelOperator
```
tsdaemon, line 15: If there was users of this operator at previously released version, they will break on update

tsdaemon (commented):
Please update CHANGES

datarobot_provider/_experimental/example_dags/model_retrain_dag.py
mjnitz02: I think this would be ok since this package isn't formally GA. I think the backwards compatibility is mainly a concern after the 11.0 GA release. Right now the package is treated as alpha/private preview

mjnitz02: I can undo the name though also, its not a huge change. I can mark it as deprecated

tsdaemon: Yeah, but before we started working on it, it was released and used by customers IIUC

paulsliusar (approved):
lgtm

PR #152, opened 2025-02-12
[MLDEV-1175] Update packaging to only include `_experimental` when building early-access version of package

c-h-russell-walker:

## Summary

Update packaging that's used in build process

## Rationale

We need and want to exclude `_experimental` from _non_ early-access builds

## Changes
- Removed usage of unsupported `python_versions` - checked [docs. here](https://setuptools.pypa.io/en/latest/references/keywords.html) after seeing warning in build output about param
- Updated to use `find_packages()` - [docs here](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages)

## Testing

Here's resulting directory output of command `make build-release` (and then the extracted version of the `.tar.gz`):

Screenshot 2025-02-12 at 6 02 46 PM

<hr>

Here's resulting directory output of command `make build-early-access` (and then the extracted version of the `.tar.gz`):

Screenshot 2025-02-12 at 6 07 39 PM


tsdaemon (approved):

paulsliusar (commented):
lgtm

paulsliusar (approved):

PR #153, opened 2025-02-13
[DM-16975] [DM-17066] Bump DataRobot client to 3.6.2 and apply relevant changes to Feature Discovery workflow

jtmcc:

## Summary
We've just release another minor patch to the python client which enables several fixes to workarounds that were originally used for the Feature Discoery workflow.
This PR bumps the client to 3.6.2 and applies the relevant Feature Discovery changes.

### DIFF
--- a/datarobot_provider/operators/feature_discovery.py
+++ b/datarobot_provider/operators/feature_discovery.py
@@ -85,7 +85,7 @@ class CreateFeatureDiscoveryRecipeOperator(BaseUseCaseEntityOperator):
                 dataset_definition["catalogVersionId"] = secondary_dataset.version_id
 
         recipe_id = recipe.id
-        recipe_config_id = recipe.settings.relationships_configuration_id
+        recipe_config_id = recipe.settings.relationships_configuration_id  # type: ignore[union-attr]
 
         # Add secondary dataset configuration information into the Recipe config
         dr.RelationshipsConfiguration(recipe_config_id).replace(
--- a/datarobot_provider/operators/feature_discovery.py
+++ b/datarobot_provider/operators/feature_discovery.py
@@ -11,7 +11,6 @@ from typing import Iterable
 from typing import Optional
 
 import datarobot as dr
-from airflow.exceptions import AirflowFailException
 from airflow.utils.context import Context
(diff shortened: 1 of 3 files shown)


demidov91 (approved):

datarobot_provider/operators/feature_discovery.py
```diff
@@ -69,29 +68,24 @@ def __init__(
...
-        response = dr.client.get_client().post(
-            "/recipes/fromDataset/",
-            data={
-                "useCaseId": self.get_use_case_id(context, required=True),
-                "status": "draft",
-                "datasetId": self.dataset_id,
-                "recipeType": "FEATURE_DISCOVERY",
-            },
+    def execute(self, context: Context) -> str:
+        use_case: dr.UseCase = self.get_use_case(context, required=True)  # type: ignore[assignment]
```
c-h-russell-walker, line 72: Sorry for the maybe nit.

IMHO when I see mypy ignores I feel they can likely be a code smell to possibly (in this case I think) be more defensive in the code.

It looks like this method `get_use_case()` can return `None`
So what about doing something like this instead to "keep mypy happy"?
```suggestion
        use_case = self.get_use_case(context, required=True)
        if not use_case:
            raise AirflowException("Use Case was not found please be sure you have defined a *use_case_id* parameter.")
```

demidov91: get_use_case(context, required=**True**) is doing what you've described above inside the method itself. So, we will never see `None` here even though the typehint is correct.

I can think of a bit different method `get_mandatory_use_case()` but should we?..

c-h-russell-walker: Oh I see - `required` is raising an exception in `get_use_case_id()` but _not_ `get_use_case()` and I think (maybe because of walrus operator?) mypy isn't smart enough to realize that.

Hmm, I guess it's just too bad if we have to have type annotation ignore statements - I think the more we can avoid them the better 🤷

demidov91: I'll make a PR to make it more mypy-friendly. I think this PR can go as-is.

tsdaemon (approved):

PR #154, opened 2025-02-13
[MLDEV-1222] Make some changes to the dev environments

mjnitz02:

## Summary
This fixes some problematic issues with the `astro` dev environment that can result from hidden volumes and other things having lingering files inside them. This should update the clean functionality to allow for a complete reset of the environment.

I am working to get a check for the environment startup into the `ci-checks`, but it has become rather difficult, so I reverted the pipeline code in this PR.

I also added some version checking tests to make sure the version ids are aligned everywhere. This was causing a side-issue where the package version and the airflow entry point receiver were not resolving properly.

This also bumps the minimum airflow version to 2.6, which was the discussed minimum supported version for now.


c-h-russell-walker (approved):
Nice updates.

docker
c-h-russell-walker, line 1: Sorry if this is a dumb question but is there a way this `docker` named file (without extension etc.) works to prunes containers that I'm unaware of?

I'd expect to see this in some bash line somewhere maybe in Makefile?

docker
```diff
@@ -0,0 +1 @@
+y container prune
```
c-h-russell-walker, line 1: Maybe easier I comment directly on the line instead of the file. 🙃

docker
c-h-russell-walker: Commented on the line `y container prune` in the file named `docker`

PR #156, opened 2025-02-14
[DM-16905] Add Feature Discovery example DAG

jtmcc:

## Summary
Add an example DAG running through all of the steps required in the Feature Discovery workbench workflow.


datarobot_provider/example_dags/feature_discovery_example.py
```diff
@@ -0,0 +1,176 @@
...
+        {"name": "enable_numeric_maximum", "value": True},
+        {"name": "enable_numeric_kurtosis", "value": False},
+        {"name": "enable_most_frequent", "value": False},
+        {"name": "enable_day", "value": True},
+        {"name": "enable_numeric_average", "value": True},
+        {"name": "enable_summarized_counts", "value": False},
+        {"name": "enable_missing_count", "value": True},
+        {"name": "enable_record_count", "value": True},
+        {"name": "enable_numeric_sum", "value": True},
+    ]
```
demidov91, line 131: Could we stick to default values in the example? That would make it more readable.

datarobot_provider/example_dags/feature_discovery_example.py
```diff
@@ -0,0 +1,176 @@
...
+        dataset2_keys=["CustomerID"],
+    )
+
+    profile_transaction_relationship = DatasetRelationshipOperator(
+        task_id="profile_transaction_relationship",
+        dataset1_identifier="profile",  # join profile
+        dataset2_identifier="transactions",  # to transactions
+        dataset1_keys=["CustomerID"],  # on CustomerID
+        dataset2_keys=["CustomerID"],
+    )
```
demidov91, line 98: We should create relationships inside the `CreateFeatureDiscoveryRecipeOperator`, there is no need to keep it as a separate action as it's not reusable.

I'm pretty sure, we're going to drop `DatasetRelationshipOperator` and `DatasetDefinitionOperator` in the `v1.0` release.

datarobot_provider/example_dags/feature_discovery_example.py
```diff
@@ -0,0 +1,176 @@
...
+        >> create_project
+        >> train_models
+        >> autopilot_complete_sensor
+    )
+
+
+datarobot_feature_discovery_pipeline_dag = datarobot_feature_discovery_pipeline()
+
+if __name__ == "__main__":
+    datarobot_feature_discovery_pipeline_dag.test()
```
demidov91, line 176:
```suggestion
datarobot_feature_discovery_pipeline()

datarobot_provider/example_dags/feature_discovery_example.py
jtmcc: Updated so that just a few are specified instead

datarobot_provider/example_dags/feature_discovery_example.py
jtmcc: > I'm pretty sure, we're going to drop DatasetRelationshipOperator and DatasetDefinitionOperato

If so then I think we should prepare for this now, rather than update later. The format of the keys expected in the input dictionaries `dataset_definitions` and `relationships` for the `CreateFeatureDiscoveryRecipeOperator` are different if they are expecting the operator **inputs**: 
[`dataset1_identifier`, `dataset1_keys`, `dataset_identifier`, etc...]

or the operator **outputs**:
[`dataset1Identifier`, `dataset1Keys`, `datasetIdentifier`, etc...]

I think it would be messy if we accepted both.. I would rather stick to always expecting to create `DatasetRelationshipOperator` and `DatasetDefinitionOperator` inside the CreateRecipe operator or never creating these mid-step operators.

I've chosen to remove entirely here, let me know if you would prefer another approach

demidov91 (approved):

PR #157, opened 2025-02-14
[MLDEV-1224] Fix input param in example dataset

mjnitz02:

## Summary
This param has to be defined as a metric for the project (e.g. RMSE, MSE, etc...). This fixes the dag so it runs in the test environments now.

We do need to probably add some better testing around this concept to ensure the example DAGs are fully working, but this is a bit of a challenging task and we're trying to sort this out as part of `MLDEV-1223` (currently in progress).


paulsliusar (approved):
lgtm

PR #158, opened 2025-02-14
[CFX-1590] Mypy rule updates and prompted related updates

c-h-russell-walker:

## Summary

Added some new mypy rules and fixed related errors

Also added more rules that we should try and enable, fix errors and then have them moving forward

## Rationale

I believe the more we can enforce mypy rules the better guardrails we'll have in place and it will also lead to more informative code as we annotate things better


bdrosen96 (approved):

mjnitz02 (approved):
Thanks for fixing these and enabling some additional mypy checks 👍

PR #164, opened 2025-02-18
[MLDEV-1228] Add xgboost version of hospitlal readmissions example dag

mjnitz02:

## Summary
This adds most of the stages for the xgboost example dag.  Currently we don't really have any way of testing dags outside of running them in a local or QA environment. This dag still needs an additional operator added for completion of everything. I'll update that in a followup.

On the side, I'm also exploring how possible it is to come up with some alternative form of testing if the dag loads properly without any errors. This would be something good to have, but I'm unsure how to do this on harness outside an active docker environment.

### DIFF
--- a/datarobot_provider/example_dags/hospital_readmissions_xgboost_example.py
+++ b/datarobot_provider/example_dags/hospital_readmissions_xgboost_example.py
@@ -1,0 +1,174 @@
+# Copyright 2025 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+
+import datarobot as dr
+from airflow.decorators import dag
+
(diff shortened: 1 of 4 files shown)


GunnerGlory (approved):

datarobot_provider/example_dags/hospital_readmissions_xgboost_example.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Upload the data into Data Registry.
+    upload_dataset = UploadDatasetOperator(task_id="upload_dataset")
+
+    # Define data preparation.
+    create_recipe = CreateWranglingRecipeOperator(
+        task_id="create_recipe",
+        dataset_id=upload_dataset.output,
+        dialect=dr.enums.DataWranglingDialect.SPARK,
+        # See the list of available *operation* options in the DtaRobot API documentation:
```
c-h-russell-walker, line 53:
```suggestion
        # See the list of available *operation* options in the DataRobot API documentation:
```

sorry for the nit - just noticed

c-h-russell-walker (approved):
Looks good!

PR #167, opened 2025-02-19
[MLDEV-1113] Add compute shap preview operator to airflow

mjnitz02:

## Summary
There isn't really a "shap" operator, so I am partially reusing the existing operator. Additionally, the calls the existing operator were using were to deprecated/no longer active API endpoints, so this operator would not work anyways.

I will create two more `Shap` operators for `distributions` and `feature impact` as well, but this is the first of the 3.


datarobot_provider/_experimental/example_dags/model_compute_shap_dag.py
mjnitz02, line 1: This dag wouldn't actually work with the existing endpoints. I'm adding SHAP to the xgboost example also as its the primary "default insights" for a model.

datarobot_provider/operators/model_insights.py
```diff
@@ -109,46 +111,44 @@ def execute(self, context: Context) -> str:
...
-        if self.model_id is None:
-            raise ValueError("model_id is required to compute SHAP impact.")
+        if not self.model_id:
+            raise AirflowFailException("The `model_id` parameter is required.")
 
     def execute(self, context: Context) -> str:
-        shap_impact_job = dr.ShapImpact.create(project_id=self.project_id, model_id=self.model_id)
-
-        self.log.info(f"Compute SHAP impact Job submitted, job_id={shap_impact_job.id}")
+        shap_preview = ShapPreview.create(entity_id=self.model_id)  # type: ignore[arg-type]
```
mjnitz02, line 152: The ignore here is a bit weird. Airflow validates through the `validate` function, but `mypy` is unaware of this call flow, and instead sees an error here. It looks like the preferred way to handle things in airflow is having default parameters be `Optional` and `validate` raises an appropriate error to the user. This sort of works against how `mypy` thinks though?

PR #170, opened 2025-02-20
[MLDEV-1232] Update docs template

mjnitz02:

## Summary
Docs for this test build can be found here: https://datarobot-datarobot-airflow-provider.readthedocs-hosted.com/en/mattn-mldev-1232-fix-docs/

I plan to deactivate this test branch set of docs and we'll move to the main branch for docs once this merges.

This will also allow us to start proofreading the docs a lot easier and make sure everything is working properly.


.readthedocs.yaml
```diff
@@ -8,15 +8,15 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
-    python: "3.13"
+    python: "3.12"
```
mjnitz02, line 11: We can't currently build on python 3.13 due to some requirements that don't support it yet.

setup.py
```diff
@@ -64,6 +64,7 @@
             "sphinx-copybutton",
             "sphinx-markdown-builder",
             "myst-parser==4.0.0",
+            "numpydoc>=1.7.0,<1.8.0",
```
mjnitz02, line 67: Needs to be a docs requirement also

c-h-russell-walker:
Glad you're getting to this - random thing I noticed in the test-build
https://datarobot-datarobot-airflow-provider.readthedocs-hosted.com/en/mattn-mldev-1232-fix-docs/CHANGES.html

The Changelog includes "Unreleased Changes" - I _feel_ like we should maybe _not_ include that in the docs. build?
Maybe easier said than done but I think that's how DR Python Client does it.

Screenshot 2025-02-20 at 2 16 43 PM

setup.py
c-h-russell-walker: Is it playing with fire to _un_-pin the ceiling for `Sphinx` in this PR now that the underlying issue was fixed?

So make: `"Sphinx>=8.1.3,<8.2.0",` into `"Sphinx>=8.1.3",` again?

mjnitz02: I can change it back if the issue was fixed and see if everything works correctly

c-h-russell-walker (approved):
Looks good 👍

PR #173, opened 2025-02-24
[MLDEV-1239] Update release pipelines for pypi and early-access pypi to have valid tokens

mjnitz02:

## Summary
Thanks to the help of IT, we now have access to both repositories. This updates the pipelines to inject the new stored secrets which are unique for each of the two repos.

Actual secrets are retrieved from the secure environments, so this is just updating the naming of how these are retrieved and changing the logic to indicate there are 2 different sets of secrets.

### DIFF
--- a/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
+++ b/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
@@ -118,4 +118,4 @@ template:
         default: test-early-access
         description: ""
         required: true
-        value: <+input>.default(test-early-access).allowedValues(test,early-access,pypi,test-early-access)
+        value: <+input>.default(test-early-access).allowedValues(release-mainline,release-early-access,test-mainline,test-early-access)
--- a/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
+++ b/.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
@@ -51,23 +51,23 @@ template:
                       REPO_URL='https://test.pypi.org/legacy/'
                       UPLOAD_SECRET='<+secrets.getValue("testpypi_token")>'
 
-                      if [ "$BUILD_TYPE" == "test-early-access" ] || [ "$BUILD_TYPE" == "early-access" ]; then
+                      if [ "$BUILD_TYPE" == "release-early-access-pypi" ] || [ "$BUILD_TYPE" == "release-early-access-test-pypi" ]; then
                           VERSION=$(python3 setup_early_access.py --version)
                       else
                           VERSION=$(python3 setup.py --version)  
                       fi
 
-                      if [ "$BUILD_TYPE" == "pypi" ]; then
+                      if [ "$BUILD_TYPE" == "release-mainline-pypi" ]; then
                           REPO_URL='https://upload.pypi.org/legacy/'
                           UPLOAD_SECRET='<+secrets.getValue("pypi_release_token")>'
-                      elif [ "$BUILD_TYPE" == "early-access" ]; then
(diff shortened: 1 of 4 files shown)


.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
```diff
@@ -106,7 +109,6 @@ template:
                       pip3 install --no-cache-dir --upgrade pip setuptools wheel six twine
                       set_build_variables
                       build_and_upload
-                      push_tags
```
mjnitz02, line 109: I'm also removing the `push_tags` concept for now. I think a separate template dedicated to github updating for tags/releases is probably better long term.

.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
```diff
@@ -57,9 +57,12 @@ template:
                           VERSION=$(python3 setup.py --version)  
                       fi
 
-                      if [ "$BUILD_TYPE" == "pypi" ] || [ "$BUILD_TYPE" == "early-access" ]; then
+                      if [ "$BUILD_TYPE" == "pypi" ]; then
```
c-h-russell-walker, line 60: I wonder if this is something I should have thought of previously but I have to think about this for a minute.

I wonder does "release-version" or "mainline-release" or something like that make more sense here?

As in we either have `"early-access"` or we have `"mainline-release"`

I'd expect to see `"pypi"` string's alternative to be `"test-pypi"`

Maybe I'm missing something though.

mjnitz02: We could change it to `mainline-release`, this was just Anton's initial suggest for the official releases to use `pypi` and `early-access`.

The 4 parameters that can be specified are:
- pypi
- early-access
- test-pypi
- test-early-access

The default selection is `test-early-access`, but I think the pipeline is always forced to select an option.

c-h-russell-walker: Hmmm, okay gotcha - I guess I think there's two things:
- Are we releasing to pypi or test-pypi
- Are we releasing early-access version or "regular" version

c-h-russell-walker: We're talking in Slack but for posterity's sake here's my suggestiong:

- release-mainline-pypi
- release-early-access-pypi
- release-mainline-testpypi
- release-early-access-testpypi

.harness/templates/publish_early_access_to_pypi_or_testpypi.yaml
```diff
@@ -51,20 +51,23 @@ template:
...
                           VERSION=$(python3 setup.py --version)  
                       fi
 
-                      if [ "$BUILD_TYPE" == "pypi" ] || [ "$BUILD_TYPE" == "early-access" ]; then
+                      if [ "$BUILD_TYPE" == "release-mainline-pypi" ]; then
                           REPO_URL='https://upload.pypi.org/legacy/'
-                          UPLOAD_SECRET='<+secrets.getValue("PyPI_token_for_airflow-provider-datarobot")>'
+                          UPLOAD_SECRET='<+secrets.getValue("pypi_release_token")>'
+                      elif [ "$BUILD_TYPE" == "release-early-access-pypi" ]; then
+                          REPO_URL='https://upload.pypi.org/legacy/'
```
c-h-russell-walker, line 64: Should this URL include `test`?

mjnitz02: Test is defined on line 51. The two `pypi` releases need to use the real upload url

c-h-russell-walker (approved):
Looks great - thanks so much for making the changes to be more verbose.

PR #175, opened 2025-02-24
[MLDEV-1131] Update early-access version to have a slightly altered dunder intit file

c-h-russell-walker:

## Summary

Early access package needs a slightly altered dunder init file since it's used as an entry point by Apache Airlfow

## Rationale

While testing the early-access provider package locally I discovered an issue where the local astro development environment would not start up.

It was caused by the underlying use of [_discover_all_providers_from_packages()](https://github.com/apache/airflow/blob/main/airflow/providers_manager.py#L588) 

And resulted in this error:
`ValueError: The package 'airflow-provider-datarobot-early-access' from setuptools and airflow-provider-datarobot do not match. Please make sure they are aligned`

Screenshot 2025-02-24 at 2 53 45 PM

## CHANGES

- Update `build-astro-dev` make command to use `build` make command
- Add new `copy-experimental-examples-astro-dev` make command to copy over relevant experimental DAGs
- Add new `build-astro-dev-early-access` make command to run local astro dev. with early-access version of pacakage
- Update mypy ignores to properly ignore deprecated but not experimental DAGs for type checking
- Update `setup_early_access.py` so that we have a dunder init file that has `"package-name": "airflow-provider-datarobot-early-access",`

### DIFF
--- a/early_access_dunder_init.txt
+++ b/early_access_dunder_init.txt
@@ -1,0 +1,69 @@
+# Copyright 2022 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from typing import TypedDict
+
+ConnectionType = TypedDict(
+    "ConnectionType",
+    {
+        "hook-class-name": str,
+        "connection-type": str,
+    },
+)
+
+
+ProviderInfoType = TypedDict(
+    "ProviderInfoType",
+    {
+        "package-name": str,
+        "name": str,
+        "description": str,
+        "versions": list[str],
+        "connection-types": list[ConnectionType],
+        "extra-links": list[str],
+    },
+)
+
+
+def get_provider_info() -> ProviderInfoType:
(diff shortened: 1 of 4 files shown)


early_access_dunder_init.txt
```diff
@@ -0,0 +1,69 @@
...
+        "extra-links": list[str],
+    },
+)
+
+
+def get_provider_info() -> ProviderInfoType:
+    return {
+        "package-name": "airflow-provider-datarobot-early-access",
+        "name": "DataRobot Airflow Provider Early Access",
+        "description": "DataRobot Airflow provider early access.",
```
c-h-russell-walker, line 36: These are the only three lines changed versus the regular dunder init

The name and description aren't necessarily needed but it's the "package-name" value that _is_ needed for Apache Airflow discovery mechanisms to work as expected.

roma-glushko (approved):

paulsliusar (approved):
lgtm

mbromanowsky (approved):
LGTM for MLDEV. This is a mostly a rubber-stamp because Matt already reviewed.

s-gavrenkov (approved):

PR #176, opened 2025-02-26
[MODEL-17080] Add StatusCheck Sensor and cleanup related docs

mjnitz02:

## Summary
Add a `StatusCheckCompleteSensor` to determine if a status check job id is completed. This is useful for determining when asynchronous tasks are completed, but allowing them to be launched in parallel.

This PR also splits the `Operators` and `Sensors` docs because they are 2 different types of airflow entities.

Sphinx is also pinned again because the `8.2.X` branch seems to be having a lot of issues, so we will just stay on `8.1.3` for now.


setup.py
```diff
@@ -57,7 +57,7 @@
             "freezegun>=1.5.1",
         ],
         "docs": [
-            "Sphinx>=8.2.0",
+            "Sphinx==8.1.3",
```
mjnitz02, line 60: Sphinx 8.2.1 is broken again.

datarobot_provider/sensors/datarobot.py
```diff
@@ -97,3 +99,52 @@ def poke(self, context: Context) -> bool:
...
+        self.job_id = job_id
+        self.datarobot_conn_id = datarobot_conn_id
+
+        self.hook = DataRobotHook(datarobot_conn_id)
+
+    def poke(self, context: Context) -> bool:
+        # Initialize DataRobot client
+        DataRobotHook(datarobot_conn_id=self.datarobot_conn_id).run()
+
+        self.log.info("Checking if scoring is complete")
```
mbromanowsky, line 136: You probably want a different log message, or none at all, for a generic status check.

mbromanowsky (approved):
LGTM for MLDEV once you get rid of the predictions-specific log message.

Some async jobs use this `StatusCheckJob` interface, while others use the `Job` class -- the difference is in the URLs, and that the `Job` class is tied to a project_id, where `StatusCheckJob` is not tied to a specific project. Are you going to need a second sensor, like `CheckProjectJobCompleteSensor`?

PR #177, opened 2025-02-27
[MMM-18684] p0 dag 2 deployment from registered model version implementation of first deployment step

paulsliusar:

## Summary
added 1st step for model deployment from model registry and its operator
added 2nd step for setting data drift settings
image

### DIFF
--- a/datarobot_provider/example_dags/hospital_readmissions_deployment_prediction_generation.py
+++ b/datarobot_provider/example_dags/hospital_readmissions_deployment_prediction_generation.py
@@ -9,6 +9,7 @@
 from airflow.decorators import dag
 
 from datarobot_provider.operators.deployment import DeployRegisteredModelOperator
+from datarobot_provider.operators.monitoring import UpdateDriftTrackingOperator
 
 """
 Example of Aiflow DAG for DataRobot data deployment and prediction generation.
@@ -17,16 +18,22 @@ Configurable parameters for this dag:
 * deployment_label - A human readable label of the deployment.
 * default_prediction_server_id - an identifier of a prediction server to be used as the default prediction server
   When working with prediction environments, default prediction server Id should not be provided
+* target_drift_enabled - if target drift tracking is to be turned on
+* feature_drift_enabled - if feature drift tracking is to be turned on
+
 """
 
 
 @dag(
     schedule=None,
+    render_template_as_native_obj=True,
     tags=["example", "csv", "predictions", "deployment"],
     params={
         "model_package_id": "",
         "deployment_label": "hospital-readmissions-example-deployment-prediction",
         "default_prediction_server_id": "",
+        "target_drift_enabled": True,
+        "feature_drift_enabled": True,
     },
 )
 def hospital_readmissions_deployment_prediction_generation():
@@ -37,7 +44,14 @@ def hospital_readmissions_deployment_prediction_generation():
         extra_params={"default_prediction_server_id": "{{ params.default_prediction_server_id }}"},
     )
 
-    deploy_registered_model
+    update_drift_tracking = UpdateDriftTrackingOperator(
+        task_id="update_drift_tracking",
+        deployment_id=deploy_registered_model.output,
+        target_drift_enabled="{{ params.target_drift_enabled }}",
+        feature_drift_enabled="{{ params.feature_drift_enabled }}",
+    )
+
+    (deploy_registered_model >> update_drift_tracking)
 
 
 hospital_readmissions_deployment_prediction_generation()
--- a/datarobot_provider/example_dags/hospital_readmissions_deployment_prediction_generation.py
+++ b/datarobot_provider/example_dags/hospital_readmissions_deployment_prediction_generation.py
@@ -1,0 +1,43 @@
+# Copyright 2025 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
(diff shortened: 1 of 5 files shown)


datarobot_provider/example_dags/hospital_readmissions_deployment_prediction_generation.py
```diff
@@ -0,0 +1,57 @@
...
+    tags=["example", "csv", "predictions", "deployment"],
+    params={
+        "model_package_id": "",
+        "deployment_label": "hospital-readmissions-example-deployment-prediction",
+        "default_prediction_server_id": "",
+        "target_drift_enabled": True,
+        "feature_drift_enabled": True,
+    },
+)
+def hospital_readmissions_deployment_prediction_generation():
```
mjnitz02, line 39: 👍 This looks good :D

datarobot_provider/operators/monitoring.py
```diff
@@ -232,3 +232,51 @@ def execute(self, context: Context) -> None:
...
+
+    This operator updates drift tracking settings for an existing deployment by calling
+    the deployment's `update_drift_tracking_settings()` method. It allows optional extra parameters
+    to be passed to the DataRobot client call.
+
+    :param deployment_id: The ID of the deployment to update.
+    :param target_drift_enabled: Boolean flag to enable target drift tracking.
+    :param feature_drift_enabled: Boolean flag to enable feature drift tracking.
+    :param kwargs: Additional keyword arguments passed to the BaseDatarobotOperator.
+    """
```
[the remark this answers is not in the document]
mjnitz02: Here's the google docstring variant

datarobot_provider/operators/deployment.py
```diff
@@ -472,3 +473,53 @@ def execute(self, context: Context) -> dict[str, Any]:
...
+    This operator creates a deployment for a registered model version by calling
+    DataRobot's `Deployment.create_from_registered_model_version()` method. It allows
+    optional extra parameters to be passed to the DataRobot client call.
+
+    :param model_package_id: The registered model version ID to deploy.
+    :param deployment_label: The label or name to assign to the deployment.
+    :param extra_params: (Optional) A dictionary of additional parameters to pass
+                         to the DataRobot deployment creation API.
+    :param kwargs: Additional keyword arguments passed to the BaseDatarobotOperator.
+    """
```
mjnitz02, line 491:
```suggestion
"""
Create a deployment from a registered model version using DataRobot's API.

This operator creates a deployment for a registered model version by calling
DataRobot's `Deployment.create_from_registered_model_version()` method. It allows
optional extra parameters to be passed to the DataRobot client call.

Args:
    model_package_id (str): The registered model version ID to deploy.
    deployment_label (str): The label or name to assign to the deployment.
    extra_params (dict, optional): A dictionary of additional parameters to pass
        to the DataRobot deployment creation API.
    kwargs (dict): Additional keyword arguments passed to the BaseDatarobotOperator.
"""
```

mjnitz02 (approved):
LGTM, left some revisions to convert the docstrings to the google docstring format

datarobot_provider/operators/deployment.py
```diff
@@ -472,3 +473,54 @@ def execute(self, context: Context) -> dict[str, Any]:
...
+        kwargs (dict): Additional keyword arguments passed to the BaseDatarobotOperator.
+    """
+
+    template_fields: Sequence[str] = ["model_package_id", "deployment_label", "extra_params"]
+
+    def __init__(
+        self,
+        *,
+        model_package_id: str,
+        deployment_label: str,
```
automactic, line 500: Why is this param having `deployment` prefix when the other operator below doesn't have `deployment` prefix, for example `deployment_target_drift_enabled`?

I personally think the param here should match exactly with the param names of `create_from_registered_model_version`, since we don't want user to second guess if `extra_params` should all have `deployment` prefix

datarobot_provider/operators/monitoring.py
```diff
@@ -232,3 +232,52 @@ def execute(self, context: Context) -> None:
...
+        self.target_drift_enabled = target_drift_enabled
+        self.feature_drift_enabled = feature_drift_enabled
+
+    def validate(self) -> None:
+        if not self.deployment_id:
+            raise ValueError("deployment_id must be provided.")
+
+    def execute(self, context: Context) -> str:
+        self.log.info("Updating drift tracking settings for deployment: %s", self.deployment_id)
+        deployment = dr.Deployment.get(self.deployment_id)
```
automactic, line 277: If you want to save a `get` call, you can probably just do `deployment = dr.Deployment(id=self.deployment_id)`, but warning I haven't actually tested this

automactic (approved):

datarobot_provider/operators/deployment.py
paulsliusar: issue was that the base class (Airflow's `BaseOperator`) had a `label` property that didn’t have a setter(its readonly for BaseOperator),which led to conflicts when i tried to assign to it.
we could override those rules with smth like this: 
```
@property
    def label(self) -> str:
        return self._label

    @label.setter
    def label(self, value: str) -> None:
        self._label = value
```
 but renaming the parameter seemed like cleaner solution

automactic: Oh in that case, having `deployment_label` does make sense

PR #183, opened 2025-03-04
[PRED-11014] Adding predictions generation operators and sensor

kessler-frost:

## Summary

Adding predictions generation airflow operators to provide similar functionality as [here](https://datarobot-public-api-client.readthedocs-hosted.com/en/latest-release/reference/predictions/predict_job.html) by the DR python SDK.


datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+        if not self.predict_job_id or not self.project_id:
+            raise AirflowFailException("predict_job_id and project_id must be provided")
+
+    def execute(self, context: Context) -> dr.PredictJob:
+        try:
+            predict_job = dr.PredictJob.get(
+                project_id=self.project_id,
+                predict_job_id=self.predict_job_id,
+            )
+        except dr.errors.PendingJobFinished:
```
andrius-senulis, line 153: In what case is this error raised?

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+    def execute(self, context: Context) -> DataFrame:
+        predictions = dr.PredictJob.get_predictions(
+            project_id=self.project_id,
+            predict_job_id=self.predict_job_id,
+        )
+
+        self.log.info(f"Predictions retrieved, predict_job_id={self.predict_job_id}")
+
+        ## NOTE: Can we even return a DataFrame here?
+        return predictions
```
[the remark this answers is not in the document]
kessler-frost: I'm not exactly sure how will this operator be used right now, but do you think it makes sense instead of s3 - for this particular case - we can instead save it as a dataset and return the id of that dataset in the project itself? this is just a dataframe right?

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+
+import datarobot as dr
+from airflow.exceptions import AirflowFailException
+from airflow.utils.context import Context
+from pandas import DataFrame
+
+from datarobot_provider.operators.base_datarobot_operator import BaseDatarobotOperator
+
+
+class GeneratePredictionsOperator(BaseDatarobotOperator):
```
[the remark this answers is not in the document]
kessler-frost: Agreed. Yeah I was going back and forth on this one, earlier (if you see the commit history) I had planned to keep it as close to our python interface as possible, but then I saw how other operators were named and made changes accordingly. I'll update the name to `RequestPredictions` 👍 .

datarobot_provider/operators/predictions.py
kessler-frost: In case the PredictJob has already finished, we don't want Airflow to keep trying to run this if the job has already finished right? And instead ask to run the relevant operator to get the predictions.

That said, we can also instead return the predictions themselves if the PredictJob has finished, but idk if that's a good way to do it since the output type will be different in different cases.

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+        except dr.errors.PendingJobFinished:
+            # NOTE: Is this the correct exception to raise in this case?
+            raise AirflowFailException(
+                "PredictJob has already finished, please use PredictJobGetPredictionsOperator to retrieve predictions"
+            ) from dr.errors.PendingJobFinished
+
+        self.log.info(f"PredictJob retrieved, predict_job_id={self.predict_job_id}")
+
+        # NOTE: Can we return the PredictJob object here?
+        return predict_job
```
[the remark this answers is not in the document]
kessler-frost: So, we are inputting the PredictJob id, and returning the same PredictJob id then?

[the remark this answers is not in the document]
kessler-frost: > So, we are inputting the PredictJob id, and returning the same PredictJob id then?

Similar to @mjnitz02 's comment. Seems a bit weird tho.

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+            project_id=self.project_id,
+            predict_job_id=self.predict_job_id,
+        )
+
+        predictions = predict_job.get_result_when_complete()
+
+        self.log.info(f"Predictions waited for and retrieved, predict_job_id={self.predict_job_id}")
+
+        # NOTE: Again, can we return a DataFrame here?
+        return predictions
```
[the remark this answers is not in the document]
kessler-frost: Sounds good, does it make sense to instead put the predictions in a dataset and return the id of that dataset? Or a similar s3 object?

datarobot_provider/operators/predictions.py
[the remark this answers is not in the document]
andrius-senulis: OK, thanks.

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+                "PredictJob has already finished, please use PredictJobGetPredictionsOperator to retrieve predictions"
+            ) from dr.errors.PendingJobFinished
+
+        self.log.info(f"PredictJob retrieved, predict_job_id={self.predict_job_id}")
+
+        # NOTE: Can we return the PredictJob object here?
+        return predict_job
+
+
+class AwaitAndGetPredictionsOperator(BaseDatarobotOperator):
```
andrius-senulis, line 165: Unfortunately, I think this operator should not exist, since it's against the Airflow design. We should have a [Sensor](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/sensors.html) which just waits for completion instead. So the flow would be this:

`RequestPredictionsOperator -> WaitForPredictionsSensor -> GetPredictionsOperator`

The `GetPredictionsOperator` should store predictions somewhere, I'm not sure where. If we want to save in Catalog, then we should have a `SavePredictionsToCatalogOperator` or smth like that.

datarobot_provider/operators/predictions.py
```diff
@@ -0,0 +1,202 @@
...
+            predict_job_id=self.predict_job_id,
+        )
+
+        self.log.info(f"Predictions retrieved, predict_job_id={self.predict_job_id}")
+
+        ## NOTE: Can we even return a DataFrame here?
+        return predictions
+
+
+class GetPredictJobOperator(BaseDatarobotOperator):
```
andrius-senulis, line 120: Unfortunately, I don't see the purpose of having this operator.

datarobot_provider/operators/predictions.py
kessler-frost: Hmm, yeah I think that makes more sense. We can use the `PredictJob.wait_for_completion` in the `WaitForPredictionsSensor` sensor.

For the "predictions storage" question, I think we have 2 options depending on our use cases:

1. Have a standard place to store the predictions ALWAYS and return some sort of id of that - that makes it easier for users to not worry about managing another resource just for their predictions
2. Write separate operators to store the predictions in separate places like s3, catalog, or some other kind of store -> I don't think that's a good idea as it will get quite complicated very quickly - so unless there's a compelling use case, we should avoid doing this

kessler-frost: I'm gonna go ahead and create that sensor and disassemble this operator.

datarobot_provider/operators/predictions.py
kessler-frost: removed this operator ✅

mjnitz02 (approved):
👍 Outside of the get predictions operator this LGTM.  I'm not 100% sure what to do with that one related to how XCOM works.

Can you add the `autodoc` stubs in `docs/autodoc_operators` and `docs/autodoc_sensors`?  You can just copy one of the existing MD files and add the references to the predictions and they'll autopopulate.

kessler-frost:
After discussion, we decided to do a "save" on the predictions' results instead of doing a "get". Thus, implementing the `SavePredictionsToDatasetOperator` operator, and the `WaitForPredictionsSensor` sensor.

PR #187, opened 2025-03-05
[MODEL-17067][MODEL-17068] Add OTV and Time Series start autopilot operator

mjnitz02:

## Summary
This PR adds a new `StartDatetimeAutopilotOperator` and splits out the actual start autopilot command to allow additional `Start*Autopilot` helper operators for other specific types of autopilots. This allows the user to be able to directly configure the operator instead of having to guess at a variety of nested dictionaries. They can still use the nested dictionary flow it they want, but this makes it a lot more straightforward.

This PR supports the following:
- OTV
- Time Series
- Unsupervised TS Clustering


datarobot_provider/operators/autopilot.py
```diff
@@ -7,34 +7,101 @@
...
+            segmentation_task_id (str, optional): The segmentation task that should be used to split the project for
+                segmented modeling.
+            max_wait_sec (int, optional): For some settings, an asynchronous task must be run to analyze the dataset.
+                max_wait governs the maximum time (in seconds) to wait before giving up.
+            datarobot_conn_id (str, optional): Connection ID, defaults to `datarobot_default`.
+        """
+
+        project = dr.Project.get(project_id)
+        if project.target:
+            raise AirflowFailException(f"Models are already trained for project_id={project.id}")
```
mariusvilkas, line 57: this probably not the case - it could be just target set?

datarobot_provider/operators/autopilot.py
```diff
@@ -7,34 +7,101 @@
...
+        if featurelist_id:
+            autopilot_settings["featurelist_id"] = featurelist_id
+
+        if relationships_configuration_id:
+            autopilot_settings["relationships_configuration_id"] = relationships_configuration_id
+
+        if segmentation_task_id:
+            autopilot_settings["segmentation_task_id"] = segmentation_task_id
+
+        autopilot_settings["max_wait"] = max_wait_sec
```
mariusvilkas, line 68: is this totally unlimited?

datarobot_provider/operators/autopilot.py
```diff
@@ -63,46 +130,239 @@ def __init__(
...
-        if (
-            "datetime_partitioning_settings" in context["params"]
-            and "partitioning_settings" in context["params"]
-        ):
-            raise ValueError(
-                "parameters: datetime_partitioning_settings and partitioning_settings are mutually exclusive"
-            )
+    def __init__(
+        self,
+        *,
```
mariusvilkas, line 241: what is the purpose of this?

datarobot_provider/operators/autopilot.py
mjnitz02: Yea, the user can technically set this to anything they want. This is the same way the python client treats it

datarobot_provider/operators/autopilot.py
mjnitz02: I thought if the target was set then that indicated that autopilot had started?

datarobot_provider/operators/autopilot.py
mjnitz02: I think it allows airflow to pass through anything. I think this actually isn't used much anymore. Let me remove it.

datarobot_provider/operators/autopilot.py
mariusvilkas: Yes, but it could be in MANUAL mode abd no "Models are already trained for project"

mariusvilkas: Maybe be something like that "Autopilot already started"?

mjnitz02: Ah, that is a really good point. I'll change the wordiing!

datarobot_provider/operators/autopilot.py
mjnitz02: Actually, weirdly it looks like this needs to be there or the operators don't construct properly. This must be related to the XCOM wrapping of the functions.

PR #189, opened 2025-03-06
[MODEL-18068] Fix some tests that don't work right on latest mac dev env

mjnitz02:

## Summary
Fixes some calls to work on mac dev environments. The `validate` operation is actually called in the context of `pre_execute` which is called by airflow, but is not actually called if the user simply calls `operator.execute()` in the context of testing. This hooks leads to some tests not working correctly in some environments. These calls should be checking the `validate` function instead.


mariusvilkas (approved):
LGTM

PR #190, opened 2025-03-06
[DM-17363] DB data prep

demidov91:
## Summary

Modify an xgboost example DAG to use a snowflake db data.

## Rationale

Users need a db data preparation example.

### DIFF
--- a/datarobot_provider/example_dags/model_training_xgboost.py
+++ b/datarobot_provider/example_dags/model_training_xgboost.py
@@ -50,8 +50,8 @@ def model_training_xgboost():
     # Create a Use Case to keep all subsequent assets. Default name is "Airflow"
     create_use_case = GetOrCreateUseCaseOperator(task_id="create_use_case", set_default=True)
 
+    # Define input data
     get_data_store = GetDataStoreOperator(task_id="get_data_store")
-
     define_transactions_table = CreateOrUpdateDataSourceOperator(
         task_id="define_transactions_table",
         data_store_id=get_data_store.output,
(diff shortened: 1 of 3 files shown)


bdrosen96 (approved):

datarobot_provider/example_dags/model_training_xgboost.py
```diff
@@ -23,40 +23,91 @@
...
     create_use_case = GetOrCreateUseCaseOperator(task_id="create_use_case", set_default=True)
 
-    # Upload the data into Data Registry.
-    upload_dataset = UploadDatasetOperator(task_id="upload_dataset")
+    # Define input data
+    get_data_store = GetDataStoreOperator(task_id="get_data_store")
+    define_transactions_table = CreateOrUpdateDataSourceOperator(
+        task_id="define_transactions_table",
+        data_store_id=get_data_store.output,
+        table_name="{{ params.secondary_table }}",
```
mjnitz02, line 58: Thanks for adding this line which assigns from the params. I think including these in the examples makes them a lot easier to read.

mjnitz02 (approved):
👍

datarobot_provider/example_dags/model_training_xgboost.py
```diff
@@ -23,40 +23,91 @@
 from datarobot_provider.sensors.model_training import ModelTrainingJobSensor
 
 """
-Example of Aiflow DAG for DataRobot data preparation and model training.
+Example of Aiflow DAG to apply db data transformations and train an xgboost model.
```
artemoboznyi, line 26: let's add a callout that in order for this dag to work as-is the user should have DR trial account

demidov91: I'll update the comment after we create those tables in the trial db.

PR #191, opened 2025-03-07
[MODEL-18048] Update pipeline ci-checks to include airflow startup testing

mjnitz02:

## Summary
This adds a new more comprehensive pipeline for testing that our package properly builds in a live airflow installation without any errors.

- Construct the test environment
- Build the latest wheel and install it into astro
- Startup the astro container
- Identify and scan the log files within the container to make sure no "Error" message are found

### DIFF
--- a/.harness/cichecks.yaml
+++ b/.harness/cichecks.yaml
@@ -161,7 +161,7 @@ pipeline:
                         shell: Bash
                         command: |-
                           sudo apt update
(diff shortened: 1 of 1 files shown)


mjnitz02:
So `grep` and `awk` just simply don't work in the harness environment properly. Piping things to them causes unknown errors on the machines with no logging. `cat` works ok, but there's not a clean way to scrape the logs with none of the standard bash operators working :/

PR #193, opened 2025-03-10
[MODEL-17043] Add lift chart, tests and docs

mjnitz02:

## Summary
This PR adds an operator to retrieve the lift chart bins, along with the associated tests and documentation additions.

### DIFF
--- a/datarobot_provider/operators/model_insights.py
+++ b/datarobot_provider/operators/model_insights.py
@@ -7,10 +7,14 @@
 # Released under the terms of DataRobot Tool and Utility Agreement.
 from collections.abc import Sequence
 from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
 
 import datarobot as dr
 from airflow.exceptions import AirflowFailException
 from airflow.utils.context import Context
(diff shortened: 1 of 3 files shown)


datarobot_provider/operators/model_insights.py
```diff
@@ -190,3 +194,55 @@ def validate(self) -> None:
...
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.project_id = str(project_id)
+        self.model_id = str(model_id)
+        self.source = str(source)
+
+    def validate(self) -> None:
+        if not self.project_id:
+            raise AirflowFailException("The `model_id` parameter is required.")
```
mariusvilkas, line 233: bad message

mariusvilkas (approved):
Similar comments like on ROC PR.
LGTM

datarobot_provider/operators/model_insights.py
mjnitz02: Ah thanks, sorry i copy pasted the actual error here

PR #194, opened 2025-03-11
[MODEL-17073] Add residuals operator, tests, and docs

mjnitz02:

## Summary
Add an operator to retrieve the residual chart values from the associated model insight.

- Operator
- Tests
- Docs


mariusvilkas (approved):
LGTM

datarobot_provider/operators/model_insights.py
```diff
@@ -302,3 +302,59 @@ def execute(self, context: Context) -> List[Dict[str, Any]]:
         model = self.get_model()
         lift_chart = model.get_lift_chart(source=self.source)
         return lift_chart.bins
+
+
+class GetResidualsChartInsightOperator(BaseDatarobotOperator):
+    """
+    Creates Residuals Chart insight data.
```
mbromanowsky, line 309:
```suggestion
    Retrieves Residuals Chart insight data.
```

datarobot_provider/operators/model_insights.py
```diff
@@ -302,3 +302,59 @@ def execute(self, context: Context) -> List[Dict[str, Any]]:
...
+        return model
+
+    def execute(self, context: Context) -> Dict[str, float]:
+        model = self.get_model()
+        residuals = model.get_residuals_chart(source=self.source)
+        return {
+            "residual_mean": residuals.residual_mean,
+            "coefficient_of_determination": residuals.coefficient_of_determination,
+            "standard_deviation": residuals.standard_deviation,
+        }
```
mbromanowsky, line 360: Is this really all we want to return? The residuals response object also has a list of chart points, and a histogram.

mbromanowsky (approved):
Reviewed for PredAI/MLDev. Fine as far as it goes, but it seems like this is leaving most of the residuals chart behind, and presenting only the three main summary statistics. If that's the intent, LGTM.

datarobot_provider/operators/model_insights.py
mjnitz02: Ah let me add that to the response as well. I think I might have missed this in the trafaret that was returned here

mjnitz02: @mbromanowsky So I went back and reviewed the portion that product wanted included, and this is all that is returned from that specific endpoint. I guess thinking about it, airflow has a very limited memory storage that can be passed between operator states, so maybe that's why this specific sdk command was selected.

I'll make a followup with product to double check this if it needs a followup 🤷

mbromanowsky: Thanks for double checking!

PR #196, opened 2025-03-12
[DM-17118] Remove outdated operators.

demidov91:
## Summary

Remove outdated Feature Discovery operators.

## Rationale

New `CreateFeatureDiscoveryRecipeOperator` creates a Feature Discovery recipe in one go. No intermediate steps needed.


alexzayats (commented):
Do we need to update documentation or somehow communicate this change to users?
I see the dag was already experimental and deprecated, but Operators were not.

alexzayats:
Fair enough, LGTM

alexzayats (approved):

mjnitz02 (approved):
Yea I think this is fine to remove. After the 11.0 release we'll have to be more careful around deprecation, but it makes sense to remove/fix old operators at this point.

PR #198, opened 2025-03-12
[MODEL-18215] Update codeowners to point to predictive-ai instead of MLDEV

mjnitz02:

## Summary
Update the codeowners based on the new org structure


CODEOWNERS
```diff
@@ -1,16 +1,16 @@
-*                                                        @datarobot/machine-learning-development
+*                                                        @datarobot/predictive-ai
```
mjnitz02, line 1: Ignore the github errors, the tags are assigned by the bots which expect these tags

akshoop (approved):

PR #201, opened 2025-03-12
[DM-17450] Remove credentials management.

demidov91:
## Summary

Remove credential management.

## Rationale

Don't try to duplicate all types of credentials DataRobot support. Just rely on the DataRobot connection name and default credentials.


datarobot_provider/operators/ai_catalog.py
```diff
@@ -229,7 +227,6 @@ def execute(self, context: Context) -> str:
         self.log.info(f"Creating Dataset from DataSource: {dataset_name}")
         ai_catalog_dataset: dr.Dataset = dr.Dataset.create_from_data_source(
             data_source_id=data_source.id,
-            credential_data=credential_data,
```
bdrosen96, line 232: Should we still support credential id at least like we do elsewhere?

demidov91: No, we shouldn't. DataRobot UI doesn't allow saving a connection without default credentials anymore. We may discuss at some point whether new endpoints should accept credential_id parameter. High-level tools like UI or Airflow should fully rely on default credentials.

bdrosen96: This is not true. What happens if user A creates a connection and has a default cred associated with it, but then shares it with user B who does not have a default cred.

demidov91: They have to copy the connection and define their own credentials.

bdrosen96: Won't this be confusing since they need to interact with two different tools?

demidov91: Update: actually, default credentials are configured per-user. So, they go to connections management, provide their credentials and work with the shared connection.

What do you mean by `two different tools`?

demidov91: The PR removes the airflow tooling for credential management to have a single source of truth. Currently the package _is trying_ to support all credential types DataRobot has. Then it duplicates the credentials in DataRobot itself. 

We can't get rid of DataRobot in the DataRobot package, so we get rid of the custom credential management.  It would be always outdated anyway because DataRobot is constantly adding new credentials types.

bdrosen96: But we still support credential_id in many places. I'm fine with not wanting to support the extendended credentials in airflow, but I think it still makes sense to allow for credential_id in all places that support it in addition to default credentials. You would still need to use the public api client or the DR ui to retrieve the credential id but it would not require duplicating credential logic in airflow since you just use it as an opaque id.

demidov91: Added

GunnerGlory (approved):

PR #203, opened 2025-03-13
Add operator to change advanced tuning parameters on models

mjnitz02:

## Summary
This PR adds an operator which allows a user to define advanced tuning parameters and apply them to a model. The operator will return the job id in the same way other model training operators work. This can be used with a `Sensor` to identify when the model has finished training.

This PR adds:
- Operator
- Operator tests
- Documentation

**NOTE: I also fixed the docstring formats for the other operators in the `model_training.py` file.**


GunnerGlory (approved):
LGTM

PR #204, opened 2025-03-14
[PRED-11176] Added Operator and DAG to run custom functions

kessler-frost:


datarobot_provider/operators/custom_function.py
```diff
@@ -0,0 +1,55 @@
...
+    Raises:
+        AirflowFailException: If custom_function is not provided
+    """
+
+    template_fields: List[str] = ["func_params"]
+
+    def __init__(
+        self,
+        *,
+        custom_func: Callable,
```
mjnitz02, line 41: Ah, I misunderstood this initially, but actually this probably works well because it gets the initial datarobot wrapping.

datarobot_provider/example_dags/custom_function_example.py
```diff
@@ -0,0 +1,74 @@
...
+        # This can be any arbitrary function
+        custom_func=post_process_predictions,
+        # The parameters must be specified here as a dictionary
+        func_params={
+            "project_id": "{{ params.project_id }}",
+            "predict_job_id": "{{ params.predict_job_id }}",
+        },
+    )
+
+    (post_process)
```
mjnitz02, line 70: After taking a second look through, I actually like this process style, and think it helps make things easier for users.

mjnitz02 (approved):
I like this approach 👍

PR #207, opened 2025-03-17
[MODEL-17053] Add CancelJobOperator

mjnitz02:

## Summary
This PR adds an operator which cancels a job is the `project_id` and `job_id` are provided. If the job has already completed, errored, or aborted, it will simply log that the job cannot be cancelled.

Tests and documentation added also.

### DIFF
--- a/datarobot_provider/operators/queue.py
+++ b/datarobot_provider/operators/queue.py
@@ -1,0 +1,60 @@
+# Copyright 2025 DataRobot, Inc. and its affiliates.
+#
+# All rights reserved.
+#
+# This is proprietary source code of DataRobot, Inc. and its affiliates.
+#
+# Released under the terms of DataRobot Tool and Utility Agreement.
+from collections.abc import Sequence
+from typing import Any
+
+import datarobot as dr
+from airflow.utils.context import Context
+from datarobot import QUEUE_STATUS
+
+from datarobot_provider.operators.base_datarobot_operator import BaseDatarobotOperator
+
+
+class CancelJobOperator(BaseDatarobotOperator):
+    """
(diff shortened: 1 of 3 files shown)


datarobot_provider/operators/queue.py
```diff
@@ -0,0 +1,60 @@
...
+    ) -> None:
+        super().__init__(**kwargs)
+        self.project_id = project_id
+        self.job_id = job_id
+
+    def validate(self):
+        if not self.project_id:
+            raise ValueError("project_id is required.")
+        if not self.job_id:
+            raise ValueError("job_id is required.")
```
arturops, line 46: This validation doesn't seem to be called inside the execute or init, maybe it is called somewhere else, but wanted to ask in case it is not called and it is important to always have project_id and job_id

arturops (approved):
Looks good to me! Very clean code :)

datarobot_provider/operators/queue.py
mjnitz02: @arturops So the `def validate` is used as a decorator by the actual calling of the airflow object when the `execute` function is triggered. The way airflow does this makes it sort of a pain for test purposes because it looks like `validate` doesn't do anything. This is formally part of the `pre_execution` functionality.

So the way airflow actually executes is:
- `pre_execute` with decorators
- `execute()`
- `post_execute` also has a set of decorators, but we don't use them

mjnitz02: During our unit testing though we just use `execute` because running the full tree is difficult to do in a unit test

arturops: Sounds great! Thanks for the detailed explanation @mjnitz02 ! 🚀

PR #231, opened 2025-04-08
[MODEL-18594] Update README.md

mjnitz02:
# This repository is public. Do not put here any private DataRobot or customer code or data.

## Summary
This PR was initially intended as just an update to the readme, but it also provides a fix for a change introduced in the datarobot 3.7 public SDK.

## Changes
- Update readme
- Fix tests/calls
- Bump compatibility for datarobot 3.7

## PR Checklist
- [ ] Changelog entry updated (`CHANGES.md`).
- [ ] Documentation added or updated (if relevant).
- [ ] Tests added or updated (if relevant).

### DIFF
--- a/datarobot_provider/operators/data_prep.py
+++ b/datarobot_provider/operators/data_prep.py
@@ -21,6 +21,7 @@ from airflow.utils.context import Context
 from datarobot import DataStore
 from datarobot.enums import DataWranglingDataSourceTypes
 from datarobot.enums import DataWranglingDialect
+from datarobot.enums import DataWranglingSnapshotPolicy
 from datarobot.enums import DownsamplingOperations
 from datarobot.enums import RecipeInputType
 from datarobot.models import DataSourceInput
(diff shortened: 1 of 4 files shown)


c-h-russell-walker (approved):

c-h-russell-walker (approved):

datarobot_provider/operators/data_prep.py
```diff
@@ -225,6 +226,7 @@ def _set_operations(self, recipe: Recipe) -> None:
                                 dataset_version_id=operation_data["arguments"][
                                     "rightDatasetVersionId"
                                 ],
+                                snapshot_policy=DataWranglingSnapshotPolicy.FIXED,
```
bdrosen96, line 229: why this change?

mjnitz02: It looks like not providing this variable when this is initialized was breaking some level of backwards compatibility. The object doesn't seem to construct correctly. This is due to this change in the public python client: https://github.com/datarobot/public_api_client/pull/3583/files

mjnitz02: The default policy now is `DataWranglingSnapshotPolicy.LATEST`. Fundamentally this doesn't really have much of an impact on this repo, but this may be something exploring more in the `public_api_client` because it might be breaking some users prior code. This is more just a side effect of the testing in this repo as it isn't GA yet.

mjnitz02: Here is another related PR for reference: https://github.com/datarobot/public_api_client/pull/3593/files

bdrosen96 (approved):

